PR #4, opened 2020-10-30
FC-790, FC-803 - Feature: S3 storage

cap10morgan:
OK, here is the big unified storage rebase that moves all (most?) of the storage functions onto the storage subsystem. This allows adding in more storage backends, starting with S3. So the changes here can be summarized as:

- All storage reads, writes, and other ops for ledger and raft go through the storage subsystem now.
- Raft logs are still local FS only because of the need for performance (and the fact that they are more local node temporal state rather than shared, persisted data).
- The storage subsystem still defaults to local FS (`:storage-type :file`).
- A new S3 storage backend is added (`:storage-type :s3`).
- Some tweaks to account for shared storage subsystems like S3 where all nodes can see the same files (e.g. only the leader writes snapshots).
- Removes ledger snapshot backend (UI removal is in FC-881 but not started yet) in anticipation of reworking that feature to be more like forking a git repo from any block (that work is also not started but captured in FC-840).
- Config params are added / tweaked to allow for all of the above.

This includes work by @zonotope too, but it all got collapsed into a big open-source repo code rebase.


bplatz:
Do we think we need the .properties settings of: fdb-storage-s3-group-prefix=group/ and fdb-storage-s3-ledger-prefix=ledger/ ? If Fluree basically owns the root path anyhow, do we need to offer the choice. Not sure if the reason was just historic, and if so perhaps we can get rid of a couple settings that are not needed.

bplatz (approved):

cap10morgan:
> Do we think we need the .properties settings of: fdb-storage-s3-group-prefix=group/ and fdb-storage-s3-ledger-prefix=ledger/ ? If Fluree basically owns the root path anyhow, do we need to offer the choice. Not sure if the reason was just historic, and if so perhaps we can get rid of a couple settings that are not needed.

I'm all for simplifying the config, especially if we can do a similar simplification in the file options too. I'll look into it. Good suggestion!

cap10morgan:
@bplatz Let me know what you think of 32ba789. I just removed the relative dir / prefix values from the sample properties file. This allows the defaults to take over in all cases. But it leaves them as an internal knob we can turn (or tell customers to turn if they bump into a need for it) for now.

bplatz:
Is there any config/.properties change someone currently running Fluree with file system storage needs to make as a result of this, or will it be backward compatible?

cap10morgan:
> Is there any config/.properties change someone currently running Fluree with file system storage needs to make as a result of this, or will it be backward compatible?

Given where we ended up I think I can make it backward-compatible pretty easily. Another commit coming shortly.

cap10morgan:
OK should be backwards compatible now.

cap10morgan:
@bplatz Is this OK to merge now or were you still doing some testing / encountering issues with it?

PR #8, opened 2020-12-06
FC-937 add jwt support to storage-handler

ldw1007:


src/fluree/db/peer/http_api.clj
```diff
@@ -827,7 +830,16 @@
                                                                                                          (if type (str "/" type))
                                                                                                          (if key (str "/" key))) host)
                                           db          (<? (fdb/db (:conn system) db-name))]
-                                      (<? (verify-auth db auth authority))))
+                                      (<? (verify-auth db auth authority)))
+
+                                    jwt
+                                    (let [jwt-auth (-> (pw-auth/fluree-auth-map (:conn system) jwt) :auth)
```
cap10morgan, line 836: This would be more idiomatic as `(:auth (pw-auth/fluree-auth-map (:conn system) jwt))` since there is only one outer fn call being threaded by `->`. Our soon-to-be-Sonarqube overlords will flag these. :)

cap10morgan (changes requested):
One idiomatic change, but LGTM o/w.

src/fluree/db/peer/http_api.clj
```diff
@@ -833,7 +833,7 @@
                                       (<? (verify-auth db auth authority)))
 
                                     jwt
-                                    (let [jwt-auth (-> (pw-auth/fluree-auth-map (:conn system) jwt) :auth)
+                                    (let [jwt-auth (:auth (pw-auth/fluree-auth-map (:conn system) jwt))
```
cap10morgan, line 836: Totally up to you, but now that I think about it, the other idiomatic way to do this would be:

```clojure
(-> system
    :conn
    (pw-auth/fluree-auth-map jwt)
    :auth)
```

cap10morgan (approved):
looks good to me!

cap10morgan (approved):
👍

PR #9, opened 2020-12-07
Use self-hosted (on AWS) GitHub Action runners

cap10morgan:
This uses AWS action runners to run our tests. Seems to be more reliable than the GitHub-provided ones. And we need to self-host Sonarqube anyway which our runners will need access to.

Those AWS runners are configured via the Terraform code in here: https://github.com/fluree/github-actions-runner-aws/pull/1

...and a GitHub app ~~that is under my GitHub account (you can't apparently create this kind of app under an organization for some reason)~~ UPDATE: I was able to transfer it to the Fluree org: https://github.com/apps/aws-action-runners

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     name: Run Tests
-    runs-on: self-hosted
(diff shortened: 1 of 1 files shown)


zonotope (approved):
🤳🏿

PR #10, opened 2021-01-08
Default to running on port 8090 to match admin UI

cap10morgan:
The admin UI defaults to looking for the ledger on port 8090, so changing the default in here to match. Anyone who has an existing config setting it to a different port should still get that after upgrading.


cap10morgan:
Sounds like this is still under discussion and we may end up defaulting everything to 8080 instead. So closing this for now.

PR #11, opened 2021-01-12
Fix/reset fulltext index workaround

zonotope:
This is a minimal quick fix for the issues around incomplete full text indexes. 

First, this patch adds more logging around fulltext index actions to make it a little easier to follow along with progress.

Next, this patch adds a "reindex-fulltext" endpoint so end users can kick off a full text reindex manually if their full text index isn't complete

This is a short term quick fix. I think fixing full text indexes so that they are more robust and resilient is a heavier lift. I've started that work on the "bug/incomplete-full-text" branches of the [ledger](https://github.com/fluree/ledger/compare/bug/incomplete-full-text) and [db](https://github.com/fluree/db/compare/bug/incomplete-full-text) repositories.

### DIFF
--- a/src/fluree/db/ledger/full_text_index.clj
+++ b/src/fluree/db/ledger/full_text_index.clj
@@ -47,32 +47,32 @@
           subject-count   (count subjects)
           timeout-ms      (or (:timeout opts) 500)
           max-retry-times (or (:retry opts) 10)]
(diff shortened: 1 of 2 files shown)


bplatz:
LGTM

bplatz (approved):

PR #12, opened 2021-01-12
Add make install target for system-wide installation

cap10morgan:
This will help us update the Homebrew formula in our tap repo.


bplatz (approved):

cap10morgan:
This addresses (but does not close) FC-980.

PR #13, opened 2021-01-12
FC-825 Update db dep

zonotope:
* Update the db dependency to pick up the updated storage path
* Adds some functions I found helpful for dealing with dev mode json files.


cap10morgan:
Just FYI, the dependency bump diff isn't showing up b/c I pushed that straight to master after rolling the release. :)

dev/user.clj
```diff
@@ -73,10 +73,15 @@
   (stop)
   (refresh-all :after 'user/re-start))
 
-(defn read-data [resource-path]
+(defn read-edn-data
+  [resource-path]
```
cap10morgan, line 77: It's dev code so not a big deal, but something like `read-edn-resource` might be more intuitive? Your change is already a big improvement, so 👍

cap10morgan (approved):
One thought on fn names, but LGTM overall!

aaj3f (approved):

PR #14, opened 2021-01-13
Use new relative block-storage-path fn

cap10morgan:
...to fix dbsync issue. Also adds some storage debug logging that I used this to debug this issue. More debug logging is always better, right?

The core of the fix is the arity change on line 233 of dbsync2.clj.


cap10morgan:
Counterpart to https://github.com/fluree/db/pull/24

cap10morgan:
This is likely the fix for FC-940.

aaj3f (approved):

bplatz (approved):

PR #15, opened 2021-01-14
FC-825 Fix/missing db chan take

zonotope:
Resets were working in a repl, but not from the http api due to a missing take from the db channel.

### DIFF
--- a/src/fluree/db/ledger/full_text_index.clj
+++ b/src/fluree/db/ledger/full_text_index.clj
@@ -49,9 +49,9 @@
           max-retry-times (or (:retry opts) 10)]
       (loop [[s & r] subjects]
         (if-not s
-          (do (when subjects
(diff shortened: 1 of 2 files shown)


ldw1007:
@zonotope is this for FC-825?

aaj3f:
> @zonotope is this for [FC-825](https://fluree.atlassian.net/browse/FC-825)?

@ldw1007 I didn't realize we had a ticket for that; this was @zonotope helping me get an intermediate solution for TrustClarity, but yes, it does solve that ticket adding a /reindex-fulltext HTTP endpoint

zonotope:
> @zonotope is this for [FC-825](https://fluree.atlassian.net/browse/FC-825)?

Yes, I just updated the title. Sorry for forgetting again.

cap10morgan (approved):
🚂

PR #16, opened 2021-01-17
FC-516 Optimize Full Text Indexing

zonotope:
This patch overhauls the full text indexing back end to eliminate race conditions and manage stateful resources more efficiently. This depends on https://github.com/fluree/db/pull/25, so we have to wait for that to be merged and a new rc published before we can merge this. CI will also fail until that happens. I've refactored the code a bit to make use of that api wrapper and to eliminate a few issues I found with the previous code. 

#### Issues with the previous approach
1. There were some [blocking takes from within a go routine](https://github.com/fluree/ledger/blob/e2be6f9e12fad0203811d8fbf091384f52effd84/src/fluree/db/ledger/full_text_index.clj#L158-L163) which could very easily [cause a deadlock](https://clojureverse.org/t/core-async-locks-up-when-using-timeout-deep-inside-async-processes/1780/9) under load. 
2. There was a [retry loop](https://github.com/fluree/ledger/blob/master/src/fluree/db/ledger/full_text_index.clj#L21-L36) with a one second back off period that was launched each time an [exception related to virtual machine locks](https://github.com/fluree/ledger/blob/e2be6f9e12fad0203811d8fbf091384f52effd84/src/fluree/db/ledger/full_text_index.clj#L69-L72) was caught. This is a [Lucene LockObtainFailedException](https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/store/LockObtainFailedException.html) that's thrown whenever a JVM tries to open multiple [IndexWriter](https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/index/IndexWriter.html) instances at once. These exceptions were popping up because previous code tries to update the index for multiple subjects at once, and all but one subject waits for a second before trying again. This effectively put a ~1 second gap between writes, and subjects in big blocks could be dropped without updating because of the retry limit.
3. There was another [lock mechanism](https://github.com/fluree/ledger/blob/master/src/fluree/db/ledger/txgroup/monitor.clj#L362-L389). While I didn't spend a lot of time diagnosing the issue because I had already decided to go with a different approach because of the problems described above, I believe this locking mechanism along with the above retry loop was causing some blocks to be dropped on the floor if they came in close behind another block that was still being processed.

#### New approach

This refactor [starts a background go routine](https://github.com/fluree/ledger/blob/fix/optimize-fulltext-index-fetch/src/fluree/db/ledger/indexing/full_text.clj#L243-L288) in a function that returns a [map of control functions](https://github.com/fluree/ledger/blob/6c72e0c41106b97efabb4899ba1c41b209881d96/src/fluree/db/ledger/indexing/full_text.clj#L287-L288) that is later [attached to the connection](https://github.com/fluree/ledger/blob/6c72e0c41106b97efabb4899ba1c41b209881d96/src/fluree/db/server.clj#L136). This background routine is responsible for [serializing full text index updates](https://github.com/fluree/ledger/blob/6c72e0c41106b97efabb4899ba1c41b209881d96/src/fluree/db/ledger/indexing/full_text.clj#L264-L265) and [managing the index writers to ensure only one is created at a time](https://github.com/fluree/ledger/blob/6c72e0c41106b97efabb4899ba1c41b209881d96/src/fluree/db/ledger/indexing/full_text.clj#L269-L270). This removes the need for both the retry loop and locking mechanism while also allowing for back pressure.

The flakes to be updated also flow through the core async channels here individually as part of a stream instead of all at once in a collection here to make it easier to serialize the index writes as well as eliminate the need for the blocking takes.


src/fluree/db/ledger/indexing/full_text.clj
```diff
@@ -0,0 +1,284 @@
...
+            [clojure.string :as str]
+            [clojure.tools.logging :as log]
+            [clucie.core :as lucene]
+            [clucie.store :as lucene-store])
+  (:import fluree.db.flake.Flake
+           java.time.Instant))
+
+(defn separate-by-op
+  [flakes]
+  (let [grouped (group-by #(.-op ^Flake %) flakes)
```
bplatz, line 16: Could destructure: (let [{add true rem false} .....

bplatz (approved):

PR #17, opened 2021-01-19
FC-996 exclude private key from raft networks section, make async

ldw1007:
Update nw-state to be 1) async and 2) use clojure-walk to remove :private-key from raft :networks section.

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -1,5 +1,6 @@
 (ns fluree.db.peer.http-api
   (:require [clojure.tools.logging :as log]
+            [clojure.walk :as walk]
             [clojure.string :as str]
             [clojure.core.async :as async]
             [aleph.http :as http]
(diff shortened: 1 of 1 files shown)


bplatz (approved):
Nice solution

src/fluree/db/peer/http_api.clj
```diff
@@ -612,69 +613,74 @@
...
+              nw-data   (remove-deep [:private-key] networks)
+              svr-state (when-let [servers (into [] (:servers leases))]
+                          (loop [[server & r] servers
+                                 acc []]
+                            (if-let [item (second server)]
+                              (recur r (into acc [{:id      (:id item)
+                                                   :active? (> (:expire item) instant)}]))
+                              acc)))
+              raft'     (-> raft
+                            (dissoc :cmd-queue :new-db-queue :networks)
```
zonotope, line 655: You can remove this line if you want because the `assoc` call below will overwrite the values for keys already present in the map

zonotope (approved):
there's one minor suggestion that you can take or leave, but looks good! 👍🏾

src/fluree/db/peer/http_api.clj
ldw1007: made changes (and verified)

ldw1007:
Made suggested changes.
Will close & merge.

PR #20, opened 2021-01-26
FC-948 fix where all flakes filtered for root auth

ldw1007:
Updates were not pushed to JS clients when a "root" auth was specified.  Flakes were filtered because of how the auth was attached to the "db".


cap10morgan (approved):
👍

PR #21, opened 2021-01-26
Fix: full-text indexer requires file storage

cap10morgan:
I ran into a test failure on master where the full-text indexer was trying to write to `/fluree` and getting a permission error as a result. It looks like this is because the tests use the memory storage option and so the file-storage-path wasn't set. The full-text indexer requires file storage (for now; changing that is a bigger lift), so this code makes it not startup if you're not using the file storage option and doesn't associate it into the conn unless it isn't nil.

Would love to get @zonotope's eyes on this especially.

### DIFF
--- a/src/fluree/db/server.clj
+++ b/src/fluree/db/server.clj
@@ -98,6 +98,13 @@
             (future
               (upgrade/upgrade conn data-version current-version))))))
 
+(defn assoc-some
+  "Assoc k -> v in m if v is not nil. Returns m unaltered otherwise."
+  [m k v]
+  (if (nil? v)
+    (assoc m k v)
+    m))
+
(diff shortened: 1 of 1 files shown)


zonotope (approved):
Nice catch, although it's weird that the failure you saw wasn't caught by ci. i see the stacktrace now, but the test still says it passes.

you're right that the full text search engine requires file storage for now, but there is an option in lucene for in-memory if we wanted it. it's worth exploring, but not now. 

in any case, lgtm

cap10morgan:
@zonotope Ah, yeah, I misspoke (miswrote?) when I said it was a test failure. More of an error log message than a test failure. Thanks for reviewing!

PR #22, opened 2021-01-28
 FC-516 Bug/empty indexer

zonotope:
The full text indexer was failing to start because we can't rely on the storage-type to be set on the connections. This patch uses the api change in https://github.com/fluree/db/pull/27 and starts the indexer if and only if the file-storage-path is set. This relies on https://github.com/fluree/db/pull/27 and can't be merged until a new library version is published.

### DIFF
--- a/src/fluree/db/server.clj
+++ b/src/fluree/db/server.clj
@@ -134,15 +134,13 @@
                                                   :memory memorystore/connection-storage-write)
                               producer-chan     (async/chan (async/sliding-buffer 100))
                               publish-fn        (local-message-process {:config config :group group} producer-chan)
-                              full-text-indexer (when (= storage-type :file) (full-text/start-indexer))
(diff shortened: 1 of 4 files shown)


src/fluree/db/server.clj
```diff
@@ -101,7 +101,7 @@
 (defn assoc-some
   "Assoc k -> v in m if v is not nil. Returns m unaltered otherwise."
   [m k v]
-  (if (nil? v)
+  (if-not (nil? v)
```
cap10morgan, line 104: D'oh! Good catch. I only did half of the logic inversion there... 🤦

cap10morgan (approved):
🧹

Thanks for sweeping up my mess!

src/fluree/db/server.clj
zonotope: heh, yeah. i wish i had caught it when i reviewed the pr before though 😉

PR #23, opened 2021-01-29
FC-1005 add newer Admin UI pages to resource list; return networks in vector

ldw1007:
A couple of changes for the latest Admin UI:
1) Added all newer pages to the http-api resource list, even the ones that previously worked
2) changed nw-state end point to return networks data in a vector.

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -642,7 +642,7 @@
                                   acc* (into acc [{(keyword k) (count v)}])]
                               (recur r acc*))
                             acc))
-              nw-data   (remove-deep [:private-key] networks)
+              nw-data   (-> (remove-deep [:private-key] networks) vector)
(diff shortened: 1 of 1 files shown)


src/fluree/db/peer/http_api.clj
```diff
@@ -642,7 +642,7 @@
                                   acc* (into acc [{(keyword k) (count v)}])]
                               (recur r acc*))
                             acc))
-              nw-data   (remove-deep [:private-key] networks)
+              nw-data   (-> (remove-deep [:private-key] networks) vector)
```
cap10morgan, line 645: This would be better written as either of the following:

`(->> networks (remove-deep [:private-key]) vector)`

or

`(vector (remove-deep [:private-key] networks))`

cap10morgan (changes requested):
One linter-level change to make. Looks good o/w.

cap10morgan (approved):
:shipit:

PR #24, opened 2021-02-04
Feature: Support homebrew 1.0 formula

cap10morgan:
This contains the changes needed for [the updated homebrew formula](https://github.com/fluree/homebrew-flureedb/pull/1) to work.

The primary change in here is the addition of the new `fdb-group-config-path` config param. This allows us to read and write files like `default-private-key.txt` when running from the Homebrew launchd plist. In this mode, there is no current directory that we can assume holds a git clone or unzipped fluree release, and we can't write to it. This config param allows me to set this to Homebrew's `etc` directory. See the linked homebrew formula PR above for details.

### DIFF
--- a/src/fluree/db/server_settings.clj
+++ b/src/fluree/db/server_settings.clj
@@ -27,6 +27,7 @@
    :fdb-consensus-type           "raft"                     ;; raft
    :fdb-encryption-secret        nil                        ;; Text encryption secret for encrypting data at rest and in transit
 
+   :fdb-group-config-path        "./"
    :fdb-group-private-key        nil
    :fdb-group-private-key-file   nil                        ;; optional location of file that contains group private key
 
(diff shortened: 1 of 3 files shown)


src/fluree/db/server_settings.clj
```diff
@@ -287,13 +288,13 @@
   (or (when-let [priv (:fdb-group-private-key env)] (str/trim priv))
       (let [priv-key-file (or (:fdb-group-private-key-file env)
                               "default-private-key.txt")
-            file          (io/file priv-key-file)
+            file          (io/file (:fdb-group-config-path env) priv-key-file)
             exists?       (.exists file)]
         (if exists?
-          (str/trim (slurp priv-key-file))
+          (str/trim (slurp file))
```
jakep36, line 294: @cap10morgan Can you explain 291 and 293 to me so I can learn? I get that you added the optional path, but why are you now slurping and spitting the file variable instead of the priv-key-file variable? Thanks

cap10morgan: @jakep36 Sure. It's because `file` contains the full path to the file (b/c of the change to line 290/291) whereas `priv-key-file` is just the filename.

jakep36: Ok. that makes sense. would it have worked to slurp file vs. priv-key-file in the existing code, too? Are they just the same in that case because they both point to a file in the current path?

jakep36 (approved):
🎠

src/fluree/db/server_settings.clj
cap10morgan: No, it wouldn't have worked. The heart of the change is line 290/291. That's what prefixes the path to the file to the filename (`clojure.java.io/file`'s signature is something akin to `(defn file [path filename] ...)`.

cap10morgan:
This is a part of (but does not complete) FC-980

PR #25, opened 2021-02-10
FC-1010, FC-748 Fix/perm check for retracted flakes

ldw1007:
Re-applying fix to automatically applying smart functions against the db-before for deletes/retractions.


src/fluree/db/ledger/transact.clj
```diff
@@ -621,7 +621,9 @@
         ;; go through each statement and check
         (loop [[^Flake flake & r] flakes]
           (when (> (.-s flake) const/$maxSystemPredicates)
-            (when-not (<? (perm-validate/allow-flake? candidate-db flake tx-permissions))
+            (when-not (if (.-op flake)
```
zonotope, line 624: The permission check here could go into its own function, but it's probably not worth it since this file will be refactored soon.

zonotope (approved):
Looks good to me ⛸️

src/fluree/db/ledger/transact.clj
ldw1007: Hopefully soon.

PR #29, opened 2021-02-16
Feature: save raft snapshot on shutdown

cap10morgan:
Requires https://github.com/fluree/raft/pull/6 (so please review that first as I need to do a release and update the dependency in here before merging). Currently I have this set to `{:local/root "../raft"}` in deps.edn, which means that the CI tests will fail until I release a new raft and update that. The tests all pass locally with that raft PR checked out at `../raft`.

This saves a final raft snapshot on shutdown so that we can restore from it later without requiring that we retain the raft logs.

There are a few other targets of opportunity in here too, but please feel encouraged to request that they be moved to a different PR if they seem controversial or worth of separate discussion.


cap10morgan:
Addresses FC-1029

bplatz (approved):

PR #32, opened 2021-02-24
Consistently use port 8090 as default everywhere

cap10morgan:
Changing the default port to 8090 slipped into master a little earlier than I intended (in some of the changes for the new Homebrew formula), so this makes that consistent everywhere. The motivation for this change is that port 8090 is _much_ more likely to be free than port 8080 on most systems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN chown -R fluree.fluree .
 USER fluree
 
 # Expose HTTP API
-EXPOSE 8080
+EXPOSE 8090
 
 # Point runtime data paths at volume
 ENV FLUREE_ARGS="-Dfdb-storage-file-root=/var/lib/fluree/"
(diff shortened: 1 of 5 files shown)


jakep36:
I found another 8080 here: https://github.com/fluree/ledger/blob/c21b7ca890fffd8f0a7a056c873ccc3bebfbe1cd/dev/user.clj#L213

cap10morgan:
> I found another 8080 here:
> 
> https://github.com/fluree/ledger/blob/c21b7ca890fffd8f0a7a056c873ccc3bebfbe1cd/dev/user.clj#L213

Yeah I left that one alone intentionally. It's setting up a query peer in an environment in which you have another node on port 8090. It's not relying on a default but explicitly setting port 8080 in development code. That all seemed OK to me.

jakep36:
👍👍

ldw1007 (approved):
It looks like Jake already approved with a thumbs up.

PR #34, opened 2021-03-01
FC-963 Maintain tspo index

zonotope:
This patch is related to https://github.com/fluree/db/pull/32/files and https://github.com/fluree/db/pull/67

The changes here include:

- Removing history nodes, keeping all history at the index nodes instead.
- Refactor the fluree.db.indexing namespace to use a depth-first, stack-based algorithm to resolve and update the index tree
- Incorporates the changes from https://github.com/fluree/db/pull/67
- Using the new `fluree.db.index/types` set in all cases where validation of index types is needed, and also maintains the new tspo index in transactions, beside other minor code refactors.

I am opening this for review for feedback for now, but I am still trying to get to the bottom of an intermittent test freeze in the [test-max](https://github.com/fluree/ledger/blob/main/test/fluree/db/ledger/docs/smart_functions/in_transactions.clj#L26-L38) smart function test. The test sometimes freezes forever when run from a terminal as part of the whole test suite, and the other times the entire test suite finishes successfully and in a reasonable amount of time. I have not gotten the test to fail or freeze at all when I run it from a repl however.


bplatz (commented):
A key part of this will be to garbage collect old block files. Don't see that here, but perhaps on the way.

cap10morgan:
@zonotope Do you still want this reviewed and merged?

zonotope:
eventually, and imminently, but it isn't ready yet. i'll convert to draft.

zonotope:
> A key part of this will be to garbage collect old block files. Don't see that here, but perhaps on the way.

Are you sure we don't want to keep the old block files around for backup and validation purposes?

bplatz:
> > A key part of this will be to garbage collect old block files. Don't see that here, but perhaps on the way.
> 
> Are you sure we don't want to keep the old block files around for backup and validation purposes?

I think instead the original transaction files now become permanently kept (with file names/keys being the txid), not block files. This provides access to the original transaction and signature data without it having to live in the Fluree indexes. Block remain but only as the source of novelty between index points, and can be garbage collected after that point.

zonotope:
> > > A key part of this will be to garbage collect old block files. Don't see that here, but perhaps on the way.
> > 
> > 
> > Are you sure we don't want to keep the old block files around for backup and validation purposes?
> 
> I think instead the original transaction files now become permanently kept (with file names/keys being the txid), not block files. This provides access to the original transaction and signature data without it having to live in the Fluree indexes. Block remain but only as the source of novelty between index points, and can be garbage collected after that point.

Are the original transaction files being purged now, and do you have a link for where the original transaction files are written?

bplatz:
> > > > A key part of this will be to garbage collect old block files. Don't see that here, but perhaps on the way.
> > > 
> > > 
> > > Are you sure we don't want to keep the old block files around for backup and validation purposes?
> > 
> > 
> > I think instead the original transaction files now become permanently kept (with file names/keys being the txid), not block files. This provides access to the original transaction and signature data without it having to live in the Fluree indexes. Block remain but only as the source of novelty between index points, and can be garbage collected after that point.
> 
> Are the original transaction files being purged now, and do you have a link for where the original transaction files are written?

Original transaction files today are never independently stored, only shared via raft (and thus stored with raft state/snapshots).
They are written to raft (and thus shared to all ledger servers) via this operation:
https://github.com/fluree/ledger/blob/1553b0d4715d7411eff4ee855a275d0325c40cee/src/fluree/db/ledger/txgroup/txgroup_proto.clj#L347

They are picked up for processing by the ledger server that is responsible for creating new blocks here: 
https://github.com/fluree/ledger/blob/1553b0d4715d7411eff4ee855a275d0325c40cee/src/fluree/db/ledger/txgroup/txgroup_proto.clj#L325

They are deleted from raft state when a new block is published. The new block contains the txids within it, and each ledger server will publish the new block and simultaneously remove the old txids from raft state here:
https://github.com/fluree/ledger/blob/1553b0d4715d7411eff4ee855a275d0325c40cee/src/fluree/db/ledger/consensus/raft.clj#L284

I think the solution is multi-step:
a) Start writing the original transaction commands via storage-write fn when published, the same way we current store block files locally. Don't yet remove it from raft state as to not break anything yet.
b) Add a new txid retrieval API/command (similar to the old block api command, that reads from disk). Now permissioned users can look up/read any full transaction data if they know the txid.
c) Stop putting _tx/tx and _tx/sig into new block flakes, these can now be retrieved via (b). Remove the current extra logic we have that keeps them out of index files (to save on size)
d) Stop putting the full transaction data into raft state since they should now be local on disk via (a). Probably need to just store the txids still in raft state, and the server that needs the full transaction (the one processing new blocks) can now read it locally from disk. Still share the full transaction via raft, we just don't store the data in raft state. This is what we do for blocks today - share the full block via raft, but the servers write it to disk, they don't put the full contents in raft state.

src/fluree/db/ledger/bootstrap.clj
```diff
@@ -222,37 +224,38 @@
...
+                                         {:status 500
+                                          :error  :db/unexpected-error})))
+         master-authid (crypto/account-id-from-message cmd sig)
+         block         (bootstrap-memory-db conn [network dbid] {:master-auth-id master-authid :txid txid :cmd cmd :sig sig})
+         new-db        (:db block)
+         block-data    (dissoc block :db)
+         _             (<? (storage/write-block conn network dbid block-data))
+         ;; todo - should create a new command to register new DB that first checks raft
+         _             (<? (txproto/register-genesis-block-async (:group conn) network dbid))
+                                        ;block-point-success? (async/<! (txproto/propose-new-block-async (:group conn) network dbid block-data))
```
dpetran, line 247: Do we need this comment? If so, do you mind changing the semicolon to `;; `, clojure-mode in emacs always tries to indent far to the right comments with a single semicolon.

src/fluree/db/ledger/indexing.clj
```diff
@@ -15,390 +17,374 @@
...
+                        :t            t
+                        :block        block
+                        :novelty-size novelty-size
+                        :start-time   start-time}]
+      (if (or (dirty? db)
+              (seq remove-preds))
+        (do (log/info "Refreshing Index:" init-stats)
+            (let [error-ch   (chan)
+                  refresh-ch (refresh-all db remove-preds error-ch)]
+              (async/alt!
```
dpetran, line 359: I like this pattern of a separate error chan and a results chan - it separates concerns nicely!

src/fluree/db/ledger/consensus/dbsync2.clj
```diff
@@ -165,14 +165,11 @@
     (loop [[leaf-key & r] leaf-keys]
       (if (nil? leaf-key)
         ::done
-        (let [leaf-his-key (str leaf-key "-his")]
-          (swap! stats-atom update :files #(+ % 2))
+        (do
+          (swap! stats-atom update :files #(+ % 1))
```
dpetran, line 169: I think this lambda in `update` could just be `inc`.

test/fluree/db/ledger/docs/query/block_query.clj
```diff
@@ -9,71 +9,88 @@
...
+
+      ;; TODO: Account for block metadata missing from the database `block`
+      ;; subject flakes for the transaction adding each block including
+      ;; `:block-bytes` and `:cmd-types`
+      (is (= (-> res keys set) #{:block :hash :instant :txns :t :sigs :flakes}))
+
+      ;; TODO: Account for filtered meta predicates which appear in block files
+      ;;       but not indexes. see fluree.db.graphdb/exclude-predicates
+      ;; (is (= 82 (count (:flakes res))))
+      (is (= 81 (count (:flakes res)))))))
```
dpetran, line 27: Just curious about the TODOs here, is this something that needs to be done eventually or will it be part of this PR?

src/fluree/db/ledger/upgrade/tspo/serde.clj
```diff
@@ -0,0 +1,102 @@
...
+  (-serialize-garbage [_ garbage]
+    (serde/serialize-garbage garbage))
+  (-deserialize-garbage [_ garbage]
+    (serde/deserialize-garbage garbage))
+  (-serialize-db-pointer [_ pointer]
+    (serde/serialize-db-pointer pointer))
+  (-deserialize-db-pointer [_ pointer]
+    (serde/deserialize-db-pointer pointer)))
+
+(defrecord IntermediateSerializer []
```
dpetran, line 77: I wouldn't mind discussing the upgrade process and how the serializer plays into that, just to make sure I understand.

zonotope: This is actually unused code that should be removed. I needed these serializers before because I thought it would be unacceptably slow to just reindex the whole db from the start. I wanted to convert the existing files in place, which would require loading the db using the old data format, and then reloading it in succession after conversion passes. Brian suggested to just reindex from the jump, which made that process a lot easier. I'll remove this stuff.

bplatz:
Looks good, let's merge!

PR #35, opened 2021-03-01
FC-789: Make adminUI a node dep

cap10morgan:
This ultimately seemed simpler than using a git submodule. Ledger doesn't need deep access to adminUI, it just needs a production build packaged at the right path, plus some dev convenience. This seems to provide all that in a way that doesn't require hairy commits & merges w/ the admin-ui build in this repo's git history.


bplatz (approved):
Great solution!

cap10morgan:
I discovered after posting this that I have some more stuff to figure out for this to work well. Namely I need to find a way to make the LESS (CSS preprocessor) pipeline support absolute import paths that reference libraries in node_modules without ejecting CRA nor using paths like `../../node_modules` (which break when admin-ui is pulled in as a dep itself). SASS supports that out of the box, but we've got a custom LESS pipeline in admin-ui. Marking this as draft until I figure out a solution.

cap10morgan:
OK along with https://github.com/fluree/fluree-admin-ui/pull/12 this seems workable now.

ldw1007 (approved):
looks good

PR #37, opened 2021-03-17
Feature: ledger version available via API

cap10morgan:
This implements FC-930 by adding a new response header `X-Fdb-Version` to every API response as well as exposing a new `/fdb/version` API resource that returns the current ledger version as plain text (should we use JSON instead? e.g. `{"version": "1.0.0-beta6"}`).

It also refactors the API and adminUI routes a bit. Let me know if you'd prefer those be split out of this PR, as I don't think they're essential to this work. I saw it more as a target-of-opportunity to clean things up in there. But if the changes are at all controversial, I can revert them for separate discussion at a later date.

This change also obviates the need to install maven, which is nice.

I'm going to build on (the approved / merged version of) this to show the current ledger version in the adminUI footer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ RELEASE_BUCKET ?= fluree-releases-public
 MINIMUM_JAVA_VERSION ?= 11
 JAVA_VERSION_FOR_RELEASE_BUILDS := $(MINIMUM_VERSION)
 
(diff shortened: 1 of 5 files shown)


ldw1007 (approved):
looks good

PR #39, opened 2021-03-26
Some improvements from the release-automation branch

cap10morgan:
- Moves group, name, and version into deps.edn and generates pom.xml on demand (should avoid the out-of-sync pom.xml issue we ran into recently)
- Generates a sha256 file with releases
- Moves print-version out of the default target spot in the Makefile so that running `make` goes back to building ledger

Even the in-progress GitHub Actions stuff should be safe to land on main. I don't think it will break anything we're currently relying on.

I'm out next week so wanted to get these in.


ldw1007 (approved):
looks good to me

PR #43, opened 2021-04-13
FC-1093 sync ws delete-ledger w/http-api version

ldw1007:
Syncing the websocket delete-ledger request with the http-api version, which actually deletes files, etc so that the ledger name can be re-used.

Also, cleaned up some code related to acquiring the connection object.


cap10morgan (approved):
🧦

PR #45, opened 2021-04-21
Fix bug in join? == true branch of launch-raft-server

cap10morgan:
I noticed this bug while working on a bigger dynamic cluster sizing improvement. Compare to line 887 for reference.


cap10morgan:
Threw in some opportunistic Makefile improvements. I noticed some targets were being run every time rather than only when needed. Figured it wasn't worth yet another PR, but let me know if you'd prefer splitting them out into their own PR.

PR #47, opened 2021-05-03
FC-1119: Run transaction tests too

cap10morgan:
Not sure why these weren't included. Ideally we wouldn't be pulling the various tests in manually like this, but instead letting the test runner find them all and run them. But at least these will get run now too.


cap10morgan:
Fixes FC-1119

zonotope (approved):
🚜

PR #53, opened 2021-06-08
FC-1036 add prop for bigdec encoding in json

ldw1007:
Should be reviewed with [PR#66](https://github.com/fluree/db/pull/66) on db. Added property to handle default behavior for bigdec encoding as non-JavaScript clients (e.g., python, c#) seem to handle BigDecimals without conversion to string.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,6 +1,6 @@
 {:deps {org.clojure/clojure {:mvn/version "1.10.3"}
         org.clojure/data.xml {:mvn/version "0.2.0-alpha6"}
-        com.fluree/db {:mvn/version "1.0.0-rc19"}
+        com.fluree/db {:mvn/version "1.0.0-rc20"}
         com.fluree/raft {:mvn/version "1.0.0-beta1"}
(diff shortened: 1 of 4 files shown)


cap10morgan (approved):
Not a blocker but just something to consider: Are there already other `fdb-json-` prefixed config params? If not, could this go under e.g. `fdb-api-` instead (maybe `fdb-api-bigdec-string`?). Just in the interest of making them intuitive and categorized.

ldw1007:
> Not a blocker but just something to consider: Are there already other `fdb-json-` prefixed config params? If not, could this go under e.g. `fdb-api-` instead (maybe `fdb-api-bigdec-string`?). Just in the interest of making them intuitive and categorized.

I am working on some other unit tests for db/ledger; so, this is definitely a do-able suggestion.  I'll include in the comment/suggestion in my current ticket.

ldw1007:
btw, unit tests should fail until the related db changes are available.  So, I do not plan to merge this branch into main until the db-library is updated.

PR #55, opened 2021-06-25
FC-1165 re-enabled close-all-sessions

ldw1007:
Re-enabled close-all-sessions when a websocket is closed with a client.


zonotope (approved):
This change looks good to me, but I think we should have a broader effort to clean up commented code. 

I think commented code should either be uncommented or removed all together. If the commented code still carries useful information, then that information should be added, in English, to the function's docstring. 

This is just my opinion, and I'm happy to hear others' ideas.

PR #56, opened 2021-06-28
FC-1162 use txMax db setting when provided

ldw1007:
Modified the select-block-commands process to use the txMax db setting when provided.  Otherwise, the default of 2mb is used for maximum transaction size.  [The default value is now defined as a constant, making it available for unit tests.]


zonotope (approved):
🔍

PR #57, opened 2021-06-29
FC-1161, FC-1168 Update/aws s3 version

ldw1007:
This merge includes the following items:

- FC-1161 Throw an exception when the _list bucket contents_ request to AWS fails
- FC-1168 Update the AWS S3 dependencies to the latest version

Note that all unit tests are passing locally.

### DIFF
--- a/src/fluree/db/ledger/storage/s3store.clj
+++ b/src/fluree/db/ledger/storage/s3store.clj
@@ -20,17 +20,23 @@
 
 (defn read
   "Returns a byte array of the data under key `k` (converted to a UNIX-style
-  path with key->unix-path) of this store's S3 bucket."
+  path with key->unix-path) of this store's S3 bucket. Returns an exception if
+  an error occurs."
   [{:keys [client bucket]} base-path k]
   (with-open [out (ByteArrayOutputStream.)]
     (let [s3-key (key->unix-path base-path k)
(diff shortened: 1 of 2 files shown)


src/fluree/db/ledger/storage/s3store.clj
```diff
@@ -94,9 +94,8 @@
                      req)
         resp       (aws/invoke client req)
         _          (when (:cognitect.anomalies/category resp)
-                     (if-let [err (:cognitect.aws.client/throwable resp)]
-                       err
-                       (ex-info "S3 list failed" {:response resp})))
+                     (throw (or (:cognitect.aws.client/throwable resp)
+                                (ex-info "S3 list failed" {:response resp}))))
```
cap10morgan, line 98: I think we probably want to return the exception instead of throwing it (like the write fn in here does). Since they're typically called async, we need to return the exception instead of throwing it.

cap10morgan (changes requested):

cap10morgan:
I pushed a commit that applies what I'm thinking to the `read` fn (that wasn't checking for errors at all before 😳).

ldw1007:
Modified list function to return an exception versus throw one. 
Also updated dependencies to latest (again).

cap10morgan (approved):
Looks great!

PR #58, opened 2021-06-30
FC-1169 fix issue where db-sync request failed for new server

ldw1007:
Fixed issue where a 'new' server could not synchronize files with an existing network.

### DIFF
--- a/src/fluree/db/ledger/consensus/raft.clj
+++ b/src/fluree/db/ledger/consensus/raft.clj
@@ -876,13 +876,13 @@
     (if join?
       ;; If joining an existing network, connects to all other servers
       (let [connect-servers (filter #(not= this-server (:server-id %)) server-configs)
-            handler-fn      (partial message-consume (:raft raft-instance) (:storage-read raft-configs))]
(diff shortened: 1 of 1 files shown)


cap10morgan (commented):
Ah, yep. Good find!

I _think_ it should maybe be `:storage-group-read` instead of `ledger`, but I'm not 100% positive. Did you try that one too?

ldw1007:
I verified that `:storage-group-read` works as well.

Should we use an 'or' condition preferring `group` over `ledger`?

cap10morgan:
> I verified that `:storage-group-read` works as well.
> 
> Should we use an 'or' condition preferring `group` over `ledger`?

No they'll generally both be set, but they point to different locations within the storage subsystem. Strange that either one would work... That makes me think there might still be an issue in here.

I'll see if I can figure out what it should be.

cap10morgan (approved):
OK I think it should be `storage-ledger-read` as you have it. I'm not totally clear on why `:storage-read` ops are handled at this layer (as well as in the state-machine), but that's orthogonal to this PR 🤷🏻‍♂️.

PR #59, opened 2021-07-01
FC-1177 Update/linting changes

ldw1007:
Last set of changes based on scanning results:

- FC-1177 Final clean-up of code per linting/scanning
- FC-1159 Configure clj-kondo scanning to ignore unused bindings for defrecord
- FC-1167 CI/CD pipeline - update dependency versions to latest possible
- FC-1160 Configure clj-kondo scanning to ignore code in comment blocks
- FC-1164 Configure clj-kondo to treat manifold.deferred/let-flow as clojure/let
- FC-1152 Remove [more] unreferenced bindings (after ignore comments...)
- FC-1163 Removed unreferenced binding http-api/sub

Some noteables:
* entire fork namespace was removed as it was not linked up anywhere other than the comment section of user namespace
* unreferenced bindings in bootstrap & raft were removed


src/fluree/db/ledger/export.clj
```diff
@@ -198,8 +196,8 @@
 
 (comment
 
-  (def db (async/<!! (fdb/db (:conn user/system) "test/one")))
-  (async/<!! (db->export db :xml))
+  (def db (clojure.core.async/<!! (fluree.db.api/db (:conn user/system) "test/one")))
+  (clojure.core.async/<!! (db->export db :xml))
```
cap10morgan, line 200: What prompted these changes? They should have been fine the way they were before. Did a linter complain for some reason?

src/fluree/db/ledger/mutable.clj
```diff
@@ -127,7 +126,7 @@
   (<?? (purge-flakes conn "fluree" "test" {:purge [422212465065991]}))
 
   (->> (<?? (storage/storage-read conn "fluree_test_block_000000000000004:v2"))
-       (serdeproto/-deserialize-block (:serializer conn)))
+       ( fluree.db.serde.protocol/-deserialize-block (:serializer conn)))
```
cap10morgan, line 129: Same question as above here. Also there's a space after the opening parens that shouldn't be there.

src/fluree/db/peer/password_auth.clj
```diff
@@ -384,7 +375,7 @@
   (let [header     {:alg "HS256",
                     :typ "JWT"}
         header-enc (-> header
-                       json/stringify-UTF8
+                       fluree.db.util.json/stringify-UTF8
```
cap10morgan, line 378: Same question here

src/fluree/db/server_settings.clj
```diff
@@ -701,7 +682,7 @@
 (comment
 
 
-  environ/env
+  environ.core/env
 
   (->
-    (build-env environ/env)))
+    (build-env environ.core/env)))
```
cap10morgan, line 688: Same question

src/fluree/db/peer/password_auth.clj
cap10morgan, line 378: Same question here

cap10morgan (commented):
I'm assuming all the code deletions are unused fns?

I'm confused by the removals of requires and then conversion of the calls to fully-qualified. They looked correct the way they were before, but even if fully qualified the namespaces should still be required (but we shouldn't do it that way).

src/fluree/db/ledger/export.clj
ldw1007: We/I turned off linting against comments since the linter(s) complained about various items (e.g., unnecessary let, unused bindings, ...)  

The above code is in the comment section.  The other option would be to delete these comment lines?

src/fluree/db/ledger/mutable.clj
ldw1007: We/I turned off linting against comments since the linter(s) complained about various items (e.g., unnecessary let, unused bindings, ...)  

The above code is in the comment section.  The other option would be to delete these comment lines?

src/fluree/db/peer/password_auth.clj
ldw1007: We/I turned off linting against comments since the linter(s) complained about various items (e.g., unnecessary let, unused bindings, ...)  

The above code is in the comment section.  The other option would be to delete these comment lines?

src/fluree/db/server_settings.clj
ldw1007: We/I turned off linting against comments since the linter(s) complained about various items (e.g., unnecessary let, unused bindings, ...)  

The above code is in the comment section.  The other option would be to delete these comment lines?

src/fluree/db/peer/password_auth.clj
ldw1007: We/I turned off linting against comments since the linter(s) complained about various items (e.g., unnecessary let, unused bindings, ...)  

The above code is in the comment section.  The other option would be to delete these comment lines?

src/fluree/db/ledger/mutable.clj
cap10morgan: Ohhh... I see. And so it still complains about unused imports then huh?

ldw1007: Yes.  I can delete code; but, I have no idea if anyone on our team still uses any of the code in the comments.  When I leave code in the comments... I am generally looking to add a unit test.

cap10morgan: Since that code is intended for REPL evaluation (I assume), we may want to just move the requires into the top of the `(comment ...)` form like this:

```clojure
(comment

  (require '[fluree.db.serde.protocol :as serdeproto])
  ...
  (serdeproto/-deserialize-block (:serializer conn)))
```

cap10morgan (approved):
I hope the linter is 😃 now!

PR #61, opened 2021-07-02
Improve :dev :test profiles in the REPL and w/r/t logging

cap10morgan:
As discussed in a recent internal Slack thread, this does the following:

1. Splits the test runner into its own `:runner` alias. `make test` now invokes `clojure -M:test:runner` to bring in both. This allows pulling in the `:test` alias in a REPL w/o also bringing in the runner.
1. Moves the `test/logback-test.xml` file to `test-resources` and only adds `test-resources` to the classpath in the `:runner` alias. This leaves the logging config alone when just using the `:test` alias without `:runner` allowing e.g. a REPL pulling in `:dev` & `:test` to use `dev/logback.xml`.

This seemed like a simple, quick starting place. Let me know if you'd like any other changes to how this works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ target/fluree-ledger.jar: resources/adminUI $(SOURCES) $(RESOURCES)
 jar: target/fluree-ledger.jar
 
 test:
-	clojure -M:test
+	clojure -M:test:runner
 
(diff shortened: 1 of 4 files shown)


bplatz (approved):

zonotope (approved):
🩺

cap10morgan:
Tests are reliably passing locally. I'm going to merge. Need to figure out why CI is so flaky though...

PR #63, opened 2021-07-16
FC-1185 update dependencies

ldw1007:
Updated dependencies related to build

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -48,7 +48,7 @@
    :main-opts ["-m" "cognitect.test-runner"]}
 
   :jar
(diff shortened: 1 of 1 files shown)


cap10morgan (changes requested):
I've found that using `RELEASE` doesn't update reliably, and there doesn't seem to be a way to request that it do so manually. It also makes it hard to know that everyone is using the same version of various tools when troubleshooting. I used to have more of these set to `RELEASE` but have been changing them to specific version numbers instead because of all of the above.

cap10morgan (approved):

PR #66, opened 2021-07-28
FC-1175: Fix S3 data recovery

cap10morgan:
This fixes the startup data recovery from snapshot when there is no more local raft state (e.g. ephemeral storage went away). Fixes FC-1175.


dpetran:
LGTM, but I don't think I want to be the final approver yet : )

zonotope (approved):
👍🏾

PR #67, opened 2021-07-30
Strip `v` prefix from docker image tags

cap10morgan:
Convention on Docker Hub is to assume these are version strings and most don't have this `v` prefix.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -44,6 +44,8 @@ jobs:
           VERSION=none
           if [[ "${{ github.event_name }}" == "schedule" ]]; then
             VERSION=nightly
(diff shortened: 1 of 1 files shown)


cap10morgan:
Requesting a review from @fluree/devrel since this mostly affects people consuming our Docker images. This changes our version tags on Docker Hub from `fluree/ledger:v1.0.0-beta13` to `fluree/ledger:1.0.0-beta13` which is more in line with how other images handle their version tags and so should be more intuitive for most people familiar with Docker Hub. But it is a change. Seemed better to get it in pre-1.0 than after.

Other non-version git tags will still be used as Docker tags as-is.

cap10morgan:
I can also push new Docker tags w/o the `v` prefix for prior releases so we can just tell everyone to use that format (rather than needing to remember when the change occurred).

PR #68, opened 2021-08-02
FC-1146 fix dropped fulltext indexes

zonotope:
Changes to use the new full-text search protocol based api defined in https://github.com/fluree/db/pull/79. These are mostly cosmetic, but many api functions from the db lib used to only take an index writer but now require passing both the index and index writer.


ldw1007 (approved):
lgtm

PR #69, opened 2021-08-05
Wrap transaction error handler in go-try

ldw1007:
The correct error is not being propagated because an async/go was missing.  
Also added a test for an expired transaction.

### DIFF
--- a/src/fluree/db/ledger/transact/error.clj
+++ b/src/fluree/db/ledger/transact/error.clj
@@ -145,19 +145,20 @@
   - transaction dependency not met
   - could not resolve auth/authority"
   [e db cmd-data t]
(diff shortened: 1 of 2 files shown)


ldw1007:
CI/CD pipeline failed with transient crypto error again:

```
ERROR in (all-open-tests auth-test createAuthority) (ECCurve.java:396)
Uncaught exception, not in assertion.
expected: nil
  actual: java.lang.IllegalArgumentException: Incorrect length for compressed encoding
 at org.bouncycastle.math.ec.ECCurve.decodePoint (ECCurve.java:396)
```

Error is unrelated to these changes.

zonotope (approved):
🥛

PR #70, opened 2021-08-11
FC-1146 destructuring cleanup; use correct reset api

zonotope:
Fixing one spot I missed with also passing the storage object to the full text api, and some minor cleanup around destructuring.


ldw1007 (approved):
lgtm

PR #72, opened 2021-08-15
FC-1210 Fix for transact error handling and db cache update

ldw1007:
fix for the following issues:

- In closed-api mode, the transaction information was not being returned.  This was caused by the db cache not being updated unless an index required an update
- In closed-api mode, query results were not updated (or were blank). This also was caused by the db cache not being updated unless an index required an update
- When a transaction error occurred (e.g., invalid auth), queries returned no results. This was caused in error handling when the returned db only contained error flakes was not converted into a 'candidate' db

### DIFF
--- a/src/fluree/db/ledger/transact.clj
+++ b/src/fluree/db/ledger/transact.clj
@@ -148,7 +148,7 @@
                                        (:dbid session) (dissoc block-result :db-before :db-after)))]
               (if (true? new-block-resp)
                 (let [res (<? (indexing/index* session {:remove-preds remove-preds*}))]
-                  (when-not res
+                  #_(when-not res
                     (let [new-db-ch (async/promise-chan)]
                       (async/put! new-db-ch (:db-after block-result))
                       (session/cas-db! session db-before-ch new-db-ch)))
--- a/src/fluree/db/ledger/transact.clj
+++ b/src/fluree/db/ledger/transact.clj
@@ -11,8 +11,7 @@
             [fluree.db.constants :as const]
             [fluree.db.util.async :refer [<? go-try]]
             [fluree.db.ledger.txgroup.txgroup-proto :as txproto]
-            [fluree.db.ledger.transact.json :as tx-json]
-            [clojure.core.async :as async]))
+            [fluree.db.ledger.transact.json :as tx-json]))
 
 
 (defn valid-authority?
@@ -147,11 +146,8 @@
                                        (-> session :conn :group) (:network session)
                                        (:dbid session) (dissoc block-result :db-before :db-after)))]
               (if (true? new-block-resp)
-                (let [res (<? (indexing/index* session {:remove-preds remove-preds*}))]
-                  #_(when-not res
-                    (let [new-db-ch (async/promise-chan)]
-                      (async/put! new-db-ch (:db-after block-result))
-                      (session/cas-db! session db-before-ch new-db-ch)))
+                (do
+                  (<? (indexing/index* session {:remove-preds remove-preds*}))
                   block-result)
                 (do
                   (log/warn "Proposed block was not accepted by the network because: "
--- a/src/fluree/db/ledger/transact.clj
+++ b/src/fluree/db/ledger/transact.clj
@@ -146,14 +146,11 @@
                                        (-> session :conn :group) (:network session)
                                        (:dbid session) (dissoc block-result :db-before :db-after)))]
               (if (true? new-block-resp)
-                (do
-                  ;; update cached db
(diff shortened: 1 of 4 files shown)


src/fluree/db/ledger/transact/error.clj
```diff
@@ -80,9 +82,11 @@
           hash-flake       (tx-meta/generate-hash-flake flakes tx-state)
           all-flakes       (conj flakes hash-flake)
           fast-forward-db? (:tt-id db-before)
-          db-after         (if fast-forward-db?
-                             (<? (dbproto/-forward-time-travel db-before all-flakes))
-                             (<? (dbproto/-with-t db-before all-flakes)))
+          db-after         (->> (if fast-forward-db?
+                                  (<? (dbproto/-forward-time-travel db-before all-flakes))
```
bplatz, line 86: Fix is to use db-root from the tx-state here. Then, you don't need to call dbproto/-rootdb. Also, there should be no reason to use tx-util/make-candidate-db - that fn is for making a proposed db before processing smartfunctions so it can clear from cache if a failure occurs. In this case, we are done processing and can proceed as we know the new db is final.

src/fluree/db/ledger/transact/error.clj
```diff
@@ -111,9 +115,11 @@
...
           all-flakes       (conj flakes hash-flake)
           fast-forward-db? (:tt-id db)
-          db-after         (if fast-forward-db?
-                             (<? (dbproto/-forward-time-travel db all-flakes))
-                             (<? (dbproto/-with-t db all-flakes)))
+          db-after         (->> (if fast-forward-db?
+                                  (<? (dbproto/-forward-time-travel db all-flakes))
+                                  (<? (dbproto/-with-t db all-flakes)))
+                                dbproto/-rootdb
+                                tx-util/make-candidate-db)
```
bplatz, line 122: Again here, use db-root from the tx-state instead of db-before (which is permissioned). Then you can remove dbproto/-rootdb and tx-util/make-candidate-db.

src/fluree/db/ledger/transact.clj
```diff
@@ -146,14 +147,11 @@
...
-                  ;  (async/put! new-db-ch (:db-after block-result))
-                  ;  (session/cas-db! session db-before-ch new-db-ch))
-                  ;; reindex if needed
-                  ;; to do -add opts
-                  (<? (indexing/index* session {:remove-preds remove-preds*}))
+                (let [res (<? (indexing/index* session {:remove-preds remove-preds*}))]
+                  (when-not res
+                    (let [new-db-ch (async/promise-chan)]
+                      (async/put! new-db-ch (:db-after block-result))
+                      (session/cas-db! session db-before-ch new-db-ch)))
```
bplatz, line 154: This code shouldn't be needed, was there an issue that was solved causing you to re-introduce this? The DB should get updated above as a result of txproto/propose-new-block-async - which blocks. This proposes the new DB to consensus, and if successful it will update the DB in the cache.

If there was an issue however we can look into fixing it, but right now this is redundant and might actually create issues.

src/fluree/db/ledger/transact/error.clj
ldw1007: This appears the be resolved for 2 or the 3 instances.  In throw-validation-exception, we are creating our own tx-state context to handle any unknowns.

zonotope (approved):
The code looks fine to me besides my minor style suggestion and Brian's comment.

src/fluree/db/ledger/transact.clj
ldw1007: This was resolved with the update to the transactor/message code

src/fluree/db/ledger/transact/error.clj
ldw1007: This was resolved with Brian's check-in.

src/fluree/db/ledger/transact/error.clj
```diff
@@ -111,9 +111,11 @@
           hash-flake       (tx-meta/generate-hash-flake flakes tx-state)
           all-flakes       (conj flakes hash-flake)
           fast-forward-db? (:tt-id db)
-          db-after         (if fast-forward-db?
-                             (<? (dbproto/-forward-time-travel db all-flakes))
-                             (<? (dbproto/-with-t db all-flakes)))
+          db-after         (->> (if fast-forward-db?
```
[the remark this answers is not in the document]
ldw1007: done.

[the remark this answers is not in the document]
ldw1007: Actually, reverted code to the original state.

zonotope (approved):

PR #74, opened 2021-08-19
FC-1223 parse flake to retrieve previous hash

ldw1007:
fixed issue in build-block where the previous hash was set a flake, versus the hash it contained.

the unit tests didn't pick this up, because in-memory db doesn't validate value-types against a map like avro does.


ldw1007:
@bplatz updated per slack conversation

bplatz (approved):

PR #75, opened 2021-08-22
FC-1231 When corrupt raft state, check for missing blocks

bplatz:
If a raft file and/or snapshot becomes corrupt and deleted, a prior state will be restored. If there were new blocks created since the last non-corrupt file, those will be on disk but not reflected in raft state.

This fix checks for newer blocks and will re-add them into raft state under the following conditions:
1) At ledger server startup only
2) When starting server is also immediately elected the leader

These conditions will always happen in a one-node ledger server, but this check will not happen if there is an existing leader.


ldw1007 (approved):
It looks intentional; but, just in case - it looks like we are:
- no longer ignoring 0-byte files
- no longer processing specific files (e.g., *.fdbd)

Otherwise, lookos good

bplatz:
@ldw1007 I moved the 0-byte and other checks into the ledger-storage namespace, so they are still there.

PR #76, opened 2021-08-24
Add direct dependency on latest aleph release

cap10morgan:
db used to bring this in, but it uses http-kit now (as ledger will soon too)


zonotope:
works for me

zonotope (approved):
⬆️

PR #77, opened 2021-08-25
FC-1235, FC-1212 return full transaction results when txid-only is false

ldw1007:
Updated command with signature to return the full transaction results when `txid-only` is false.  [If the transaction wait expires, a 408 will be returned.]

The following defaults are used so as not to introduce breaking changes.  
- For command w/signature: `txid-only` defaults to true.
- For command without signature: previously `txid-only` was hard-coded to false.  [Customer/Partner request.]


ldw1007:
Added unit tests

PR #80, opened 2021-08-27
FC-1239 Fix/tx ex reporting

bplatz:
This refactors block building to be a bit more broken out. Along the way many more exception conditions are handled to keep the transactor loop running smoothly.

A new capability to remove a transaction from the queue in raft state is added, such that exceptions can not only report out but remove the offending transaction(s) and continue moving forward.

This addresses tickets FC-1239 and FC-1242

### DIFF
--- a/test/fluree/db/ledger/transact/core_test.clj
+++ b/test/fluree/db/ledger/transact/core_test.clj
@@ -2,9 +2,8 @@
   (:require [clojure.test :refer :all]
             [clojure.core.async :as async]
             [fluree.db.ledger.memorydb :as memorydb]
-            [fluree.db.ledger.transact.core :as tjson]
-            [fluree.db.api :as fdb]
-            [fluree.db.util.json :as json])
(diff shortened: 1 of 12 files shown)


test/fluree/db/ledger/transact/core_test.clj
```diff
@@ -2,7 +2,7 @@
   (:require [clojure.test :refer :all]
             [clojure.core.async :as async]
             [fluree.db.ledger.memorydb :as memorydb]
-            [fluree.db.ledger.transact.json :as tjson]
+            [fluree.db.ledger.transact.core :as tjson]
```
ldw1007, line 5: not a blocker, we can change alias when we add unit tests.

ldw1007 (approved):
lgtm

test/fluree/db/ledger/transact/core_test.clj
bplatz: I'll fix... no tests in there quite yet.

PR #81, opened 2021-08-27
FC-1216 Fix/tx unique swap

bplatz:
This fix addresses error that could occur if swapping unique predicate IDs across two subjects in the same transaction as described by FC-1216.

Fix involves (a) adjusting validation function so it never returns nil, and returns true on success, and (b) adding a nil guard to put result on a channel - in case any other future functions accidentally return nil values.


ldw1007 (approved):
lgtm

adding a unit test or maybe a generic unit test ticket for all of the changes we recently made?

PR #82, opened 2021-09-01
Make reflection fail CI builds

cap10morgan:
It's going to be hard to keep reflection out of our code unless we have tools that can help us catch it early and often.

This PR does two things:
1. Adds a new `make ci` target that runs our test suite and the eastwood linter (which catches reflection, among many other useful things; it caught a bug the first time I ran it).
2. Runs the `make ci` target in our CI pipeline so that things like reflection will fail the build.

If this pattern looks good to folks I'll apply it to other repos too.

This builds on #79 so it will look pretty gnarly until that's merged. Marking as draft until then.

### DIFF
--- a/src/fluree/db/ledger/consensus/raft.clj
+++ b/src/fluree/db/ledger/consensus/raft.clj
@@ -22,8 +22,6 @@
 
 (set! *warn-on-reflection* true)
 
-(set! *warn-on-reflection* true)
-
 
 (defn snapshot-xfer
   "Transfers snapshot from this server as leader, to a follower.
--- a/src/fluree/db/ledger/consensus/raft.clj
+++ b/src/fluree/db/ledger/consensus/raft.clj
@@ -22,6 +22,8 @@
 
 (set! *warn-on-reflection* true)
 
+(set! *warn-on-reflection* true)
+
 
 (defn snapshot-xfer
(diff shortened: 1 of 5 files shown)


cap10morgan:
OK this is ready for review. @ldw1007 in particular I wanted to make sure 38b5d7f doesn't screw up anything for you.

src/fluree/db/ledger/consensus/raft.clj
```diff
@@ -22,6 +22,8 @@
 
 (set! *warn-on-reflection* true)
 
+(set! *warn-on-reflection* true)
```
zonotope, line 25: Looks like an extra line here

src/fluree/db/ledger/indexing.clj
```diff
@@ -15,6 +15,8 @@
 
 (set! *warn-on-reflection* true)
 
+(set! *warn-on-reflection* true)
```
zonotope, line 18: another extra line

zonotope (approved):
🤠 (<---- Clint Eastwood)

ldw1007:
This looks good to me.  We were using a file versus standard output for the eastwood results for a client's scanning script.  We can work with them (when they are ready), to adjust their scanning script.

ldw1007 (approved):

PR #84, opened 2021-09-03
Remove unused ledger-port config param

cap10morgan:
...and add fdb-group-port to master list of config params w/ an
explanation of what it does when.


Tr3yb0 (commented):
Do we need to list group-port in the sample properties file?

cap10morgan:
> Do we need to list group-port in the sample properties file?

I don't think so. It shouldn't typically be used. But I will if others disagree. :)

Tr3yb0 (approved):

PR #86, opened 2021-09-09
FC-1096 Added information about pending transactions to nw-state

ldw1007:
* Added pending transaction information to nw-state endpoint
* Moved nw-state-handler and health-handler into a new namespace, server-health.  If you can think of a better name.
* Added unit tests for nw-state-handler and health-handler - named server-health-tests so it is executed only by the ledger-test entry point (for now).
* seemed odd that the ledger-test and test-helper namespaces were not in the fluree.db.peer tree; so, I moved "ledger" from their namespace

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -1,8 +1,7 @@
 (ns fluree.db.peer.http-api
   (:require [clojure.tools.logging :as log]
-            [clojure.walk :as walk]
             [clojure.string :as str]
-            [clojure.core.async :as async :refer [<!!]]
+            [clojure.core.async :as async]
             [org.httpkit.server :as http]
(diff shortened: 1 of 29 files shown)


ldw1007:
@cap10morgan You may be interested in this one, as I tweaked the names of several namespaces for unit tests.

cap10morgan:
> @cap10morgan You may be interested in this one, as I tweaked the names of several namespaces for unit tests.

Yeah, those changes are fine. We have a ticket for a much bigger test layout refactoring, so this will all change much more soon (I hope).

cap10morgan (approved):
Looks good. I like the tests!

I think for namespace moves w/ changes like this (and this is as much a note to self as anyone else), we should try to remember to move the existing code to the new ns first (in one commit) and then make the changes after that (in one or more subsequent commits). It's hard to tell what all changed otherwise.

PR #88, opened 2021-09-14
Lint test code too

cap10morgan:
I had eastwood scan the test code in db and it found several good things to fix, so I thought I'd apply the same change here. And sure enough it found several things to fix in here too.

Please take careful note of the changes to some of the tests that weren't being run / checked. I basically just made them reflect the actual values being returned now. I looked around a little in the code / commit history but that didn't really clarify it.

Part of FC-899.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ test:
 	clojure -X:test
 
 eastwood:
-	clojure -M:eastwood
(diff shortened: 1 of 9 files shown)


ldw1007 (approved):
lgtm

PR #90, opened 2021-09-20
Add test dir to dev alias

cap10morgan:
This allows you to run tests (e.g. from a REPL) with just the dev alias
loaded, and doesn't bring in test-resources in case you want to see the
log output. If you don't, you can bring in the test alias too and
logging will be disabled.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -38,7 +38,7 @@
   :mvn/version "1.0.0-beta15"
 
   :dev
-  {:extra-paths ["dev"]
+  {:extra-paths ["dev", "test"]
    :extra-deps  {org.clojure/tools.namespace {:mvn/version "1.1.0"}}}
 
   :test


zonotope (approved):
I have this in my branches. I should have issued a tiny pr doing this a while ago. My bad.

dpetran:
While we're at it, can we add `"src-cljs"` and `"src-nodejs`" to the `:dev` `:extra-paths`? That's something that I've got locally so that I can use a browser/node cljs repl.

cap10morgan:
> While we're at it, can we add `"src-cljs"` and `"src-nodejs`" to the `:dev` `:extra-paths`? That's something that I've got locally so that I can use a browser/node cljs repl.

That would go in db, not ledger. :) But sure, I don't see why not.

PR #92, opened 2021-09-22
FC-1254, FC-1270 Fix/permissioned crawls

bplatz:
This adds new tests to ensure the fixes fluree/db are tested as it relates to leaking out 'nil' values for unpermissioned data instead of not listing any values.


test/fluree/db/ledger/general/todo_permissions.clj
```diff
@@ -127,7 +149,7 @@
     (is (= 5 (:block td-resp)))
 
     ;; there should be 4 tempids
```
ldw1007, line 151: Note to self (more than anyone): remove comment because check is self-explanatory.  We can do that the next time we are updating tests.

ldw1007 (approved):
lgtm

zonotope (approved):

ldw1007 (approved):

PR #93, opened 2021-09-23
FC-1265: Fix invalid full text predicate in older ledgers

zonotope:
It looks like some ledgers populated before the full text index tracking feature was released either didn't have or didn't use a migration to add the [reserved predicate](https://github.com/fluree/db/blob/main/src/fluree/db/constants.cljc#L55). Instead, at least for Ledgerset, an ad-hoc predicate was added that used the `_predicate/fullText` name, but had a different subject id from the reserved one. Since the predicate is [recognized by it's subject id alone](https://github.com/fluree/db/blob/main/src/fluree/db/constants.cljc#L55), no data was tracked in the full text index for those ledgers.

This patch adds a namespace to retract the reserved `_predicate/fullText` predicate name from the add-hock predicates, if they exist, transacts the correct, reserved predicate with the correct id, updates all the existing data to use the new (reserved) predicate, and then resets the full text index.

I kept running into race conditions where the database returned from `fluree.db.api/db` would not reflect the data that was just transacted, so I have a (kinda) dirty [loop](https://github.com/fluree/ledger/compare/fix/invalid-fulltext-subid?expand=1#diff-5bfcbef41460ad4b316b6d1a11f8f0bb7f023dcf318ef2190816606ade4b1434R17-R23) to wait until the next block. I would love a suggestion for a better way to do that without needing a `Thread/sleep` call.


src/fluree/db/ledger/upgrade/full_text.clj
```diff
@@ -0,0 +1,120 @@
...
+  (:import (fluree.db.flake Flake)))
+
+(set! *warn-on-reflection* true)
+
+(def pred-name "_predicate/fullText")
+(def invalid-name "_predicate/fullText-invalid")
+
+(defn wait-for-block
+  [conn ledger expected-block]
+  (loop [{:keys [block] :as db} (<!! (fdb/db conn ledger))]
```
ldw1007, line 19: @bplatz looking for other options than a Thread/sleep loop.  

- would sync-to be an option? We have the expected block.
- or would it better to use monitor-tx (with a very long expiration) for the last submitted transaction?

bplatz: You can use: `(<!! (fdb/db conn ledger {:syncTo expected-block}))`

ldw1007 (approved):
lgtm

PR #94, opened 2021-09-28
FC-1274 Fix issue for signed requests where digest is not calculated on the original body content

ldw1007:
Occasionally, the calculated digest did not match the digest provided in the header.   The body of a request first was converted from bytes into a string and then parsed into JSON.   The body was then re-stringified to calculate the digest.  The JSON->stringify operation may or may not retain the original order of contents.  

This impacted the following endpoints:
- fdb/delete-db
- fdb/network/db/{action} (e.g., fdb/myNetwork/ledgerOne/query)

With this fix, the digest is calculated on the original (stringified) body contents, prior to its conversion into JSON. 

Also, I added a test for a signed-multi-query that supplies its map contents in non-alphanumeric order.  This unit test failed prior to the changes. 

The unit test for the `delete-db ` endpoint will be submitted with another PR.  A change in processing was required to support an in-memory data store.  However, you cannot validate the digest calculation for delete-db without a closed api.

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -112,10 +112,10 @@
 
 (defn decode-body
   [body type]
-  (let [^bytes body' (.bytes ^BytesInputStream body)]
+  (let [body' (-> ^BytesInputStream body .bytes (String. "UTF-8"))]
     (case type
-      :string (String. body' "UTF-8")
-      :json   (-> body' (String. "UTF-8") json/parse))))
+      :string body'
+      :json   (json/parse body'))))
 
 
 (defn- return-token
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -112,9 +112,10 @@
 
 (defn decode-body
   [body type]
-  (case type
-    :json (let [^bytes body' (.bytes ^BytesInputStream body)]
-            (-> body' (String. "UTF-8") json/parse))))
+  (let [^bytes body' (.bytes ^BytesInputStream body)]
+    (case type
+      :string (String. body' "UTF-8")
+      :json   (-> body' (String. "UTF-8") json/parse))))
 
 
 (defn- return-token
@@ -485,8 +486,9 @@
         start           (System/nanoTime)
         ledger          (keyword network db)
         action*         (keyword action)
-        action-param    (when body (decode-body body :json))
-        auth-map        (auth-map system ledger request action-param)
+        body'           (when body (decode-body body :string))
+        action-param    (some-> body' json/parse)
+        auth-map        (auth-map system ledger request body')
         request-timeout (if-let [timeout (:request-timeout headers)]
                           (try (Integer/parseInt timeout)
                                (catch Exception _ 60000))
@@ -660,11 +662,12 @@
 
 (defn delete-ledger
   [{:keys [conn] :as system} {:keys [body remote-addr] :as request}]
-  (let [body         (when body (decode-body body :json))
-        ledger-ident (:db/id body)
+  (let [body-str     (when body (decode-body body :string))
+        body'        (some-> body-str json/parse)
+        ledger-ident (:db/id body')
         [nw db]      (str/split ledger-ident #"/")
         ledger       (keyword nw db)
-        auth-map     (auth-map system ledger request body)
+        auth-map     (auth-map system ledger request body-str)
         session      (session/session conn [nw db])
         db*          (<?? (session/current-db session))
         ;; TODO - root role just checks if the auth has a role with id 'root' this can
--- a/test/fluree/db/ledger/api/downloaded.clj
+++ b/test/fluree/db/ledger/api/downloaded.clj
@@ -182,9 +182,14 @@
       (is (every? boolean (map #(predicates %) ["comment/nestedComponent" "person/stringUnique"]))))))
 
 (deftest sign-multi-query
-  (testing "sign-multi-query where collections are not named in alphanumeric order"
+  (testing "sign multi-query where collections are not named in alphanumeric order"
     (let [private-key   (slurp "default-private-key.txt")
-          qry-str       "{\"collections\":{\"select\":[\"*\"],\"from\":\"_collection\"},\n       \"predicates\":{\"select\":[\"*\"],\"from\":\"_predicate\"},\n       \"_setting\":{\"select\":[\"*\"],\"from\":\"_setting\"},\n       \"_rule\":{\"select\":[\"*\"],\"from\":\"_rule\"},\n       \"_role\":{\"select\":[\"*\"],\"from\":\"_role\"},\n       \"_user\":{\"select\":[\"*\"],\"from\":\"_user\"}\n      }"
+          qry-str       (str "{\"collections\":{\"select\":[\"*\"],\"from\":\"_collection\"},\n "
+                             " \"predicates\":{\"select\":[\"*\"],\"from\":\"_predicate\"},\n  "
+                             " \"_setting\":{\"select\":[\"*\"],\"from\":\"_setting\"},\n "
+                             " \"_rule\":{\"select\":[\"*\"],\"from\":\"_rule\"},\n "
+                             " \"_role\":{\"select\":[\"*\"],\"from\":\"_role\"},\n "
+                             " \"_user\":{\"select\":[\"*\"],\"from\":\"_user\"}\n }")
           request       {:headers {"content-type" "application/json"}
                          :body    qry-str}
           q-endpoint    (str endpoint-url "multi-query")
@@ -201,14 +206,14 @@
       ; The keys in the response are -> :opts :body :headers :status
       (is (test/contains-many? resp :opts :body :headers :status))
 
-      ; Are all the predicates what we expect?
-      (is (= collections #{"_rule" "nestedComponent" "_fn" "_predicate" "_setting" "chat" "_auth" "_user" "person" "_shard" "_tag" "comment" "_role" "_collection"}))
+      ; Are all the collections what we expect?
+      (is (test/contains-many? collections "_rule" "nestedComponent" "_fn" "_predicate" "_setting" "chat" "_auth" "_user" "person" "_shard" "_tag" "comment" "_role" "_collection"))
 
       ; Are some of the predicates we expect returned?
-      (is (every? boolean (map #(predicates %) ["comment/nestedComponent" "person/stringUnique"])))
+      (is (test/contains-many? predicates "comment/nestedComponent" "person/stringUnique"))
 
       ; Are the expected roles returned?
-      (is (= roles #{"chatUser" "root"})))))
+      (is (test/contains-many? roles "chatUser" "root")))))
 
 (deftest query-collections-predicates*
   (add-schema*)
@@ -297,22 +302,28 @@
 (deftest sign-all-collections-graphql
   (testing "sign a query for all collections through the graphql endpoint"
     (let [private-key         (slurp "default-private-key.txt")
-          qry-str             (json/stringify {:query "{  graph {  _collection (sort: {predicate: \"name\", order: ASC}) { _id name spec version doc}}}"})
+          graphql-str         (str "{  graph {  _collection "
+                                   "(sort: {predicate: \"name\", order: ASC})"
+                                   "{ _id name spec version doc}}}")
+          qry-str             (json/stringify {:query graphql-str})
           request             {:headers {"content-type" "application/json"}
                                :body    qry-str}
           q-endpoint          (str endpoint-url "graphql")
           signed-req          (http-signatures/sign-request :post q-endpoint request private-key)
           resp                @(http/post q-endpoint signed-req)
           body                (-> resp :body bs/to-string json/parse)
-          collection-names    (into #{} (map #(:name %) (-> body :data :_collection)))]
+          collections         (-> body :data :_collection)
+          collection-keys     (reduce-kv
+                                (fn [result _ collection] (->> collection keys (into result)))
+                                #{}
+                                collections)
+          collection-names    (into #{} (map #(:name %) collections))]
 
-      ; Are the collection keys what we expect?
-      (is (map #(#{:doc :version :spec :name :_id} %) body))
+      ; Are the keys in the collections what we expect?
+      (is (test/contains-many? collection-keys :_id :name :version :doc))
 
       ;; Are the collections what we expect?
-      (is (not (empty? (remove nil? (map #((into #{} collection-names) %) ["_rule" "_fn" "_predicate" "_setting" "_auth" "_user" "_shard" "_tag" "_role" "_collection"])))))
-
-      (is (every? boolean (map #(collection-names %) ["_rule" "_fn" "_predicate" "_setting" "_auth" "_user" "_shard" "_tag" "_role" "_collection"]))))))
+      (is (test/contains-many? collection-names "_rule" "_fn" "_predicate" "_setting" "_auth" "_user" "_shard" "_tag" "_role" "_collection")))))
 
 
 (deftest repeated-query-all-collections-graphql*
@@ -390,7 +401,8 @@
       ;; Each result should be an array of 1 (?name)
       (is (every? boolean (map #(= 1 (count %)) body)))
 
-      (is (every? boolean (map #(collection-names %) ["_predicate" "_auth" "_collection" "_fn" "_role" "_rule" "_setting" "_tag" "_user"]))))))
+      (is (test/contains-many? collection-names
+                               "_predicate" "_auth" "_collection" "_fn" "_role" "_rule" "_setting" "_tag" "_user")))))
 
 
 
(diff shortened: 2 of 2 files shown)


ldw1007:
We picked up support for signed sql, graphQL and SPARQL queries with this change.

src/fluree/db/peer/http_api.clj
```diff
@@ -112,9 +112,10 @@
 
 (defn decode-body
   [body type]
-  (case type
-    :json (let [^bytes body' (.bytes ^BytesInputStream body)]
-            (-> body' (String. "UTF-8") json/parse))))
+  (let [^bytes body' (.bytes ^BytesInputStream body)]
+    (case type
+      :string (String. body' "UTF-8")
+      :json   (-> body' (String. "UTF-8") json/parse))))
```
cap10morgan, line 118: If we want the string conversion in both cases it might make sense to do that up in the let form?

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -191,6 +192,36 @@
...
+          status       (:status resp)
+          body         (some-> resp :body bs/to-string json/parse)
+          collections  (into #{} (map #(:_collection/name %) (:collections body)))
+          predicates   (into #{} (map #(:_predicate/name %) (:predicates body)))
+          roles        (into #{} (map #(:_role/id %) (:_role body)))]
+
+      (is (= 200 status))
+
+      ; The keys in the response are -> :request-time :aleph/keep-alive? :headers :status :connection-time :body
+      (is (not (empty? (remove nil? (map #(#{:request-time :aleph/keep-alive? :headers :status :connection-time :body} %)
```
cap10morgan, line 214: If you want to test that all of these keys are in responseKeys, you could do `(is (every? (map #(#{...}) responseKeys)))`.

Also if we're not interoperating with an environment that prefers different idioms (e.g. Java / JavaScript), `response-keys` would be a better name.

And are you sure that `:aleph/keep-alive?` could be in there? We're not using aleph anymore.

cap10morgan (commented):
Some thoughts / questions, but looking good overall 😁

test/fluree/db/ledger/api/downloaded.clj
ldw1007: :aleph/keep-alive -> we may need to look at all of the unit tests

src/fluree/db/peer/http_api.clj
ldw1007: ✅

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -84,9 +84,8 @@
 
       (is (= 200 status))
 
-      ; The keys in the response are -> :request-time :aleph/keep-alive? :headers :status :connection-time :body
-      (is (not (empty? (remove nil? (map #(#{:request-time :aleph/keep-alive? :headers :status :connection-time :body} %)
-                                         response-keys)))))
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (every? (map #(#{:opts :body :headers :status})) response-keys))
```
bplatz, line 88: I think this would be:
(is (every? (map #{:opts :body :headers :status} response-keys)))

test/fluree/db/ledger/api/downloaded.clj
ldw1007: Made following changes across unit tests:
- renamed responseKeys to response-keys
- when aleph was mentioned, updated response-keys based on http-kit
- above let to a few formatting changes

test/fluree/db/ledger/api/downloaded.clj
bplatz: In this case using a set as a function (http://blog.jayfields.com/2010/08/clojure-using-sets-and-maps-as.html)

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -121,16 +120,15 @@
...
           body                (-> queryCollectionsRes :body bs/to-string json/parse)
           collections         (into #{} (map #(:_collection/name %) body))]
 
       (is (= 200 status))
 
-      ; The keys in the response are -> :request-time :aleph/keep-alive? :headers :status :connection-time :body
-      (is (not (empty? (remove nil? (map #(#{:request-time :aleph/keep-alive? :headers :status :connection-time :body} %)
-                                         responseKeys)))))
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (every? (map #(#{:opts :body :headers :status})) response-keys))
```
bplatz, line 131: Same problem as above

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -148,17 +146,16 @@
...
           predicates          (into #{} (map #(:_predicate/name %) body))]
 
 
       (is (= 200 status))
 
-      ; The keys in the response are -> :request-time :aleph/keep-alive? :headers :status :connection-time :body
-      (is (not (empty? (remove nil? (map #(#{:request-time :aleph/keep-alive? :headers :status :connection-time :body} %)
-                                         responseKeys)))))
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (every? (map #(#{:opts :body :headers :status})) response-keys))
```
bplatz, line 158: Ditto

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -171,20 +168,19 @@
...
+          collections   (into #{} (map #(:_collection/name %) (:coll body)))
+          predicates    (into #{} (map #(:_predicate/name %) (:pred body)))]
 
       (is (= 200 status))
 
-      ; The keys in the response are -> :request-time :aleph/keep-alive? :headers :status :connection-time :body
-      (is (not (empty? (remove nil? (map #(#{:request-time :aleph/keep-alive? :headers :status :connection-time :body} %)
-                                         responseKeys)))))
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (every? (map #(#{:opts :body :headers :status})) response-keys))
```
bplatz, line 183: Ditto

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -194,25 +190,24 @@
...
+          predicates    (into #{} (map #(:_predicate/name %) (:predicates body)))
+          roles         (into #{} (map #(:_role/id %) (:_role body)))]
 
       (is (= 200 status))
 
-      ; The keys in the response are -> :request-time :aleph/keep-alive? :headers :status :connection-time :body
-      (is (not (empty? (remove nil? (map #(#{:request-time :aleph/keep-alive? :headers :status :connection-time :body} %)
-                                         responseKeys)))))
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (every? (map #(#{:opts :body :headers :status})) response-keys))
```
bplatz, line 210: Ditto

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -427,22 +422,21 @@
...
+          status        (:status resp)
+          body          (some-> resp :body bs/to-string json/parse)
+          collections   (into #{} (map #(:_collection/name %) body))]
       (is (= 200 status))
 
-      ; The keys in the response are -> :request-time :aleph/keep-alive? :headers :status :connection-time :body
-      (is (not (empty? (remove nil? (map #(#{:request-time :aleph/keep-alive? :headers :status :connection-time :body} %)
-                                         responseKeys)))))
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (every? (map #(#{:opts :body :headers :status})) response-keys))
```
bplatz, line 439: Ditto

test/fluree/db/ledger/api/downloaded.clj
cap10morgan: ah, yep. no need to wrap the set in a fn literal in that context. 👍🏻

cap10morgan: Yeah, that makes sense. These map key set checks are always slightly less intuitive to me than they seem. And in fact, when I actually threw it a REPL, it was `(every? #{:opts :body :headers :status} response-keys)` w/o the `map`. But what that checks (as you saw) is that every key in the map is one of the keys in the set literal. It doesn't care if one or more of the keys in the set is missing from the map!

Whereas the body of `contains-many?` is just `(every? #(contains? m %) ks)` which is saying "for every key in `ks`, check that the map `m` contains it" and is really what we want here. Good fix!

test/fluree/db/ledger/api/downloaded.clj
ldw1007: using test/contains-many?

test/fluree/db/ledger/api/downloaded.clj
ldw1007: using test/contains-many?

test/fluree/db/ledger/api/downloaded.clj
ldw1007: using test/contains-many?

test/fluree/db/ledger/api/downloaded.clj
ldw1007: using test/contains-many?

test/fluree/db/ledger/api/downloaded.clj
ldw1007: As noted below, we used `text/contains-many?` to check the response keys.

ldw1007:
@bplatz and @cap10morgan
Take a look one more time?

src/fluree/db/peer/http_api.clj
```diff
@@ -112,9 +112,10 @@
 
 (defn decode-body
   [body type]
-  (case type
-    :json (let [^bytes body' (.bytes ^BytesInputStream body)]
-            (-> body' (String. "UTF-8") json/parse))))
+  (let [body' (-> ^BytesInputStream body .bytes (String. "UTF-8"))]
```
cap10morgan, line 115: This might cause reflection w/o the `^bytes` type hint. Did you run `make eastwood` on it?

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -170,27 +162,53 @@
...
       ; Are all the predicates what we expect?
       (is (= collections #{"_rule" "nestedComponent" "_fn" "_predicate" "_setting" "chat" "_auth" "_user" "person" "_shard" "_tag" "comment" "_role" "_collection"}))
 
       ; Are some of the predicates we expect returned?
       (is (every? boolean (map #(predicates %) ["comment/nestedComponent" "person/stringUnique"]))))))
 
+(deftest sign-multi-query
+  (testing "sign-multi-query where collections are not named in alphanumeric order"
+    (let [private-key   (slurp "default-private-key.txt")
+          qry-str       "{\"collections\":{\"select\":[\"*\"],\"from\":\"_collection\"},\n       \"predicates\":{\"select\":[\"*\"],\"from\":\"_predicate\"},\n       \"_setting\":{\"select\":[\"*\"],\"from\":\"_setting\"},\n       \"_rule\":{\"select\":[\"*\"],\"from\":\"_rule\"},\n       \"_role\":{\"select\":[\"*\"],\"from\":\"_role\"},\n       \"_user\":{\"select\":[\"*\"],\"from\":\"_user\"}\n      }"
```
cap10morgan, line 187: Might be nicer to break this string up into multiple lines (e.g. with `str`)

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -170,27 +162,53 @@
...
+      (is (= 200 status))
+
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (test/contains-many? resp :opts :body :headers :status))
+
+      ; Are all the predicates what we expect?
+      (is (= collections #{"_rule" "nestedComponent" "_fn" "_predicate" "_setting" "chat" "_auth" "_user" "person" "_shard" "_tag" "comment" "_role" "_collection"}))
+
+      ; Are some of the predicates we expect returned?
+      (is (every? boolean (map #(predicates %) ["comment/nestedComponent" "person/stringUnique"])))
```
cap10morgan, line 208: I think this can be simplified to:

`(is (every? predicates ["comment/nestedComponent" "person/stringUnique"]))` since any missing result `nil` should cause `every?` to return `false`.

Or you could just use `contains-many?` here too since all of the provided keys should be in the set (which also support `contains?` on values so you can pass the set as the `m` arg to `contains-many?` and the vec of string literals as `ks`).

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -170,27 +162,53 @@
...
+          collections   (into #{} (map #(:_collection/name %) (:collections body)))
+          predicates    (into #{} (map #(:_predicate/name %) (:predicates body)))
+          roles         (into #{} (map #(:_role/id %) (:_role body)))]
+
+      (is (= 200 status))
+
+      ; The keys in the response are -> :opts :body :headers :status
+      (is (test/contains-many? resp :opts :body :headers :status))
+
+      ; Are all the predicates what we expect?
```
cap10morgan, line 204: s/predicates/collections/ ?

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -276,6 +294,26 @@
...
+          request             {:headers {"content-type" "application/json"}
+                               :body    qry-str}
+          q-endpoint          (str endpoint-url "graphql")
+          signed-req          (http-signatures/sign-request :post q-endpoint request private-key)
+          resp                @(http/post q-endpoint signed-req)
+          body                (-> resp :body bs/to-string json/parse)
+          collection-names    (into #{} (map #(:name %) (-> body :data :_collection)))]
+
+      ; Are the collection keys what we expect?
+      (is (map #(#{:doc :version :spec :name :_id} %) body))
```
cap10morgan, line 310: Another good candidate for `contains-many?`?

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -276,6 +294,26 @@
...
+
+      ;; Are the collections what we expect?
+      (is (not (empty? (remove nil? (map #((into #{} collection-names) %) ["_rule" "_fn" "_predicate" "_setting" "_auth" "_user" "_shard" "_tag" "_role" "_collection"])))))
```
cap10morgan, line 313: `contains-many?`?

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -276,6 +294,26 @@
...
+
+      (is (every? boolean (map #(collection-names %) ["_rule" "_fn" "_predicate" "_setting" "_auth" "_user" "_shard" "_tag" "_role" "_collection"]))))))
```
cap10morgan, line 315: Another `contains-many?` candidate

cap10morgan (changes requested):
Getting there. Still some test code that could be simplified or in some cases I think isn't quite the right logic as-is. The `(every? boolean ...)` forms are good candidates for simplification (some or all with `contains-many?`), for example. I didn't exhaustively comment on all of them below, but the others are hopefully easy to spot.

src/fluree/db/peer/http_api.clj
ldw1007: It passed the `make ci` without any issues.

I'll test with the following:
```
(defn decode-body
  [^bytes body type]
  (let [body' (-> ^BytesInputStream body .bytes (String. "UTF-8"))]
    (case type
      :string body'
      :json   (json/parse body'))))
```
^^^
That passes linting as well

bplatz (approved):

src/fluree/db/peer/http_api.clj
cap10morgan: If the original code doesn't bother the linter then it's 👍🏻

test/fluree/db/ledger/api/downloaded.clj
ldw1007: ✅

test/fluree/db/ledger/api/downloaded.clj
ldw1007: ✅

test/fluree/db/ledger/api/downloaded.clj
ldw1007: ✅

test/fluree/db/ledger/api/downloaded.clj
ldw1007: ✅

test/fluree/db/ledger/api/downloaded.clj
```diff
@@ -297,22 +302,28 @@
...
           signed-req          (http-signatures/sign-request :post q-endpoint request private-key)
           resp                @(http/post q-endpoint signed-req)
           body                (-> resp :body bs/to-string json/parse)
-          collection-names    (into #{} (map #(:name %) (-> body :data :_collection)))]
+          collections         (-> body :data :_collection)
+          collection-keys     (reduce-kv
+                                (fn [result _ collection] (->> collection keys (into result)))
+                                #{}
+                                collections)
+          collection-names    (into #{} (map #(:name %) collections))]
```
cap10morgan, line 320: Not a big deal but this can be `(set (map :name collections))`

cap10morgan (approved):
Looks good!

PR #95, opened 2021-09-28
FC-1280 only invoke full-text indexer when process is defined on the connection

ldw1007:
While writing a unit test for delete ledger to validate changes on another ticket, I ran into an issue where the full-text indexer process was not defined. This is because the unit tests use in-memory storage for our data. 

Modified the delete-ledger code to not invoke the indexer process, unless it is defined.  This modification allows a `delete-db ` endpoint unit test to be included in our ci/cd.

Manually verified that full-text directories were cleaned up when submitting a _signed_ `delete-db` request in closed-api mode.


zonotope (approved):
☑️

PR #96, opened 2021-10-05
FC-1275 (i guess): Dev utility functions

zonotope:
The code here consists of a few functions I found useful while debugging the wildcard query issues reported in FC-1275. The fix turned out to be fully contained in the db library and there are no user-facing changes here, but I think these dev-mode functions would be useful for me and future developers in fluree core going forward. Hence this pr.

### DIFF
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -107,17 +107,6 @@
   (Thread/sleep 1000)
   (load-ledger-resources db-name loader resource-paths))
 
-(defn query-db [db-name query]
-  @(-> system
-       :conn
-       (fdb/db db-name)
-       (fdb/query query)))
-
-(defn sql-query-db [db-name sql-query]
-  (->> sql-query
-       sql/parse
-       (query-db db-name)))
-
 (defn load-chat-ledger
   []
   (let [collection-txn  [{:_id  "_collection"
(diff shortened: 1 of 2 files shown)


dev/user.clj
```diff
@@ -117,6 +118,41 @@
...
+        predicate-txn   (edn/read-string (slurp (io/resource pred-filename)))
+
+        data-filename   "../test/fluree/db/ledger/Resources/ChatApp/chatAppData.edn"
+        data-txn        (edn/read-string (slurp (io/resource data-filename)))]
+    (log/info "Creating new ledger")
+    @(fdb/new-ledger (:conn system) test-helpers/ledger-chat)
+    (Thread/sleep 250)
+
+    (log/info "Adding collections")
+    (async/<!! (fdb/transact-async (:conn system) test-helpers/ledger-chat collection-txn))
```
bplatz, line 144: Unless you are working in a go-loop, I prefer the non-asyc API:
`@(fdb/transact (:conn system) test-helpers/ledger-chat collection-txn)`

dev/user.clj
```diff
@@ -117,6 +118,41 @@
...
+    (Thread/sleep 250)
+
+    (log/info "Adding predicates")
+    (async/<!! (fdb/transact-async (:conn system) test-helpers/ledger-chat predicate-txn))
+    (Thread/sleep 250)
```
bplatz, line 149: I don't think this sleep would be required here at all. The one after @(fdb/new-ledger ...) is the only one that should be a got'cha, along with possibly queries immediately after a transaction (but this can be controlled with the :syncTo option).

If the sleep here is needed between transactions, I'd consider that a bug.

bplatz (commented):
Good idea to get this in there, I have to do this all the time and end up duplicating my work.

PR #109, opened 2021-11-10
FC-1305: Full Text Startup Sync

zonotope:
The full text indexer connected to connection objects under the `:fulltext/indexer` key is not a function but a map containing a processor function under the `:process` key which adds new items to the index by adding those items to the indexing queue, and a function under the `:close` key that closes the indexing queue. We previously were treating the indexer itself as the processor function instead of the actual correct processor function.

This patch first calls the full text indexer correctly by calling the function under the `:process` key instead of the full text indexer itself. Next, the raft startup function then synchronizes the full text indexer with the current state of the database, and will terminate the startup process if there are any unexpected errors during sync.


ldw1007 (approved):
lgtm

PR #110, opened 2021-11-17
FC-1305: don't try to sync the index if all blocks are registered

zonotope:
My previous patch caused an error to show up in the logs during dbsync when the full text index is up to date due to block-range calls where the start block ended up being greater than the end block. Block range happily returns a reversed sequence of blocks in this case, but the indexer throws the error when trying to index blocks out of order. This does not affect ledger startup as the blocks are already indexed, but this patch ensures the error isn't thrown in this case to clean up the logs. It does this by short circuiting sync if the index is up to date.


cap10morgan (approved):
🧇

PR #111, opened 2021-11-18
Fix/query peer work

dpetran:
This is, if not the final word, then at least an improvement on how the query server works. 

Issues specifically addressed:
- now we operate a query group properly
- we have a basic test framework to assure us of some minimal amount of functionality

### DIFF
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -215,6 +215,17 @@
 
   (async/poll! vres))
 
+(defn start-one
+  ([] (start {}))
+  ([override-settings]
+   (let [settings (-> (settings/build-env environ/env)
+                      (merge override-settings))]
+     (server/startup settings))))
+
+
+(defn stop-one [s]
+  (when s (server/shutdown s))
+  :stopped)
 
 
 
@@ -251,6 +262,25 @@
           :fdb-consensus-type "in-memory"})
 
   (stop)
+
+  (def ledger-peer (start-one {:fdb-api-port 8090
+                               :fdb-mode "ledger"
+                               :fdb-group-servers "ledger-server@localhost:11001"
+                               :fdb-group-this-server "ledger-server"
+                               :fdb-group-log-directory "./build/data/group"
+                               :fdb-storage-file-root "./build/data"
+                               #_#_:fdb-storage-type "memory"
+                               #_#_:fdb-consensus-type "in-memory"}))
+  (def query-peer (start-one {:fdb-api-port 8099
+                              :fdb-mode "query"
+                              :fdb-query-peer-servers (str "localhost:" 8090)
+                              :fdb-group-servers "query-server@localhost:11002"
+                              :fdb-group-this-server "query-server"
+                              :fdb-storage-type "memory"
+                              :fdb-consensus-type "in-memory"}))
+
+  (stop-one ledger-peer)
+  (stop-one query-peer)
 
 
   ;; Three servers
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -253,31 +253,31 @@
 
 
   ;; Query Peer
-  (start {:fdb-api-port 8080
-          :fdb-mode "query"
+  (start {:fdb-api-port           8080
+          :fdb-mode               "query"
           :fdb-query-peer-servers "localhost:8090"
-          :fdb-group-servers "query-server@localhost:11002"
-          :fdb-group-this-server "query-server"
-          :fdb-storage-type "memory"
-          :fdb-consensus-type "in-memory"})
+          :fdb-group-servers      "query-server@localhost:11002"
+          :fdb-group-this-server  "query-server"
+          :fdb-storage-type       "memory"
+          :fdb-consensus-type     "in-memory"})
 
   (stop)
 
-  (def ledger-peer (start-one {:fdb-api-port 8090
-                               :fdb-mode "ledger"
-                               :fdb-group-servers "ledger-server@localhost:11001"
-                               :fdb-group-this-server "ledger-server"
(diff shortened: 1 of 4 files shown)


dev/user.clj
```diff
@@ -242,12 +253,35 @@
...
 
+  (def ledger-peer (start-one {:fdb-api-port 8090
+                               :fdb-mode "ledger"
+                               :fdb-group-servers "ledger-server@localhost:11001"
+                               :fdb-group-this-server "ledger-server"
+                               :fdb-group-log-directory "./build/data/group"
+                               :fdb-storage-file-root "./build/data"
+                               #_#_:fdb-storage-type "memory"
+                               #_#_:fdb-consensus-type "in-memory"}))
+  (def query-peer (start-one {:fdb-api-port 8080
```
Tr3yb0, line 274: should we change this port to something other than 8080, for the same reasons we moved the ledger port to 8090?

dpetran: This is just for local development so I don't think it's too big of a deal, but I can change it just to be safe.

dev/user.clj
```diff
@@ -242,12 +253,35 @@
...
-  (start {:fdb-query-peer-servers "localhost:8090"
-          :fdb-api-port            8080
-          :fdb-mode                "query"})
+  (start {:fdb-api-port 8080
+          :fdb-mode "query"
+          :fdb-query-peer-servers "localhost:8090"
+          :fdb-group-servers "query-server@localhost:11002"
+          :fdb-group-this-server "query-server"
+          :fdb-storage-type "memory"
+          :fdb-consensus-type "in-memory"})
```
cap10morgan, line 262: Can you align those k-v's please? (something something still need to add a .editorconfig one of these days haha)

dev/user.clj
```diff
@@ -242,12 +253,35 @@
...
   (stop)
 
+  (def ledger-peer (start-one {:fdb-api-port 8090
+                               :fdb-mode "ledger"
+                               :fdb-group-servers "ledger-server@localhost:11001"
+                               :fdb-group-this-server "ledger-server"
+                               :fdb-group-log-directory "./build/data/group"
+                               :fdb-storage-file-root "./build/data"
+                               #_#_:fdb-storage-type "memory"
+                               #_#_:fdb-consensus-type "in-memory"}))
```
cap10morgan, line 273: Do we still want those last two?

src/fluree/db/server_settings.clj
```diff
@@ -445,39 +445,37 @@
...
+                           :novelty-max       (-> settings :fdb-memory-reindex-max env-bytes)
+                           :dev?              dev?
+                           :password-auth     (password-feature-settings settings)
+                           :open-api          (-> settings :fdb-api-open env-boolean)
+                           :file-storage-path (when (= :file storage-type)
+                                                file-ledger-storage-path)
+                           :s3-storage        (when (= :s3 storage-type)
+                                                {:bucket (:bucket s3-conn)
+                                                 :prefix s3-ledger-storage-prefix})
+                           :encryption-secret encryption-key}}}))
```
cap10morgan, line 478: Needs some alignment here too

test/fluree/db/peer/query_peer_tests.clj
```diff
@@ -0,0 +1,92 @@
...
+                    :body
+                    json/parse
+                    (map #(dissoc % :_id)))
+               (->> query-query-resp
+                    :body
+                    json/parse
+                    (map #(dissoc % :_id)))))))
+
+    (stop query-peer)
+    (stop ledger-peer)))
```
cap10morgan, line 92: Yay tests! 🎉 🌮

cap10morgan (approved):
Just some minor formatting stuff and one question. LGTM once addressed!

dev/user.clj
dpetran: I found it handy to be able to switch back-and-forth between in-memory and persisted. Usually I'm against leaving commented out code like this (that's what git is for), but in the `dev/user` ns it seemed more convenient. I'm happy to drop it you think we ought, though.

cap10morgan: Totally fine in dev/user

test/fluree/db/peer/query_peer_tests.clj
dpetran: Can you think of any reason running multiple servers on the same JVM would be a bad idea? It's super convenient here, but I'm not 100% sure that it's okay.

src/fluree/db/server_settings.clj
dpetran: I feel a little silly but I don't see what's unaligned here. Happy to change it, but I'm not sure what to change : )

test/fluree/db/peer/query_peer_tests.clj
cap10morgan: Seems OK to me! But yeah, there may be dragons in there...

src/fluree/db/server_settings.clj
cap10morgan: Sorry, it's up above here. I just scrolled to the bottom of the whole map to make my comment.

dpetran: Oh wait, I see that you must mean the whole `generate-conn-settings` function body.

cap10morgan: ...and in doing so scrolled to the next file hahaha. It's in dev/user actually. My bad!

cap10morgan: Oh wait, GitHub was hiding things from me again. Yes, I mean this file but lines 450-468. Sorry I'm making this extra confusing now!

PR #115, opened 2021-12-07
Fix/unresolved remote addr

dpetran:
This is something I cooked up with @jakep36 to solve the issue laid out in https://fluree.atlassian.net/browse/FC-1320

The gist of the problem is that if a remote server in the `:fdb-group-servers` config has an unresolvable hostname, then the tcp thread crashes when we try to connect to it. This fixes that by just spinning until the remote host is resolveable.

This is the simplest fix that we could come up with that addresses the issue, but that doesn't mean there can't be better approaches, so advice is welcome there.

Another question is if we should make this configurable - I don't know why we wouldn't want this behavior, but maybe we want to configure the retry strategy in some way. Feedback welcome.

### DIFF
--- a/src/fluree/db/ledger/consensus/tcp.clj
+++ b/src/fluree/db/ledger/consensus/tcp.clj
@@ -4,7 +4,7 @@
             [taoensso.nippy :as nippy]
             [fluree.db.util.async :refer [go-try]]
             [clojure.tools.logging :as log])
-  (:import (java.net BindException)))
+  (:import (java.net BindException InetSocketAddress)))
 
 (set! *warn-on-reflection* true)
 
@@ -29,9 +29,9 @@
         (recur)))
     ;; return shutdown function
     (fn []
+      (log/info "Shutting down tcp server on port" port)
       (async/close! accept-chan)
       (ntcp/shutdown! evt-loop))))
-
 
 ;;; implementation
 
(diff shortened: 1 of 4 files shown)


jakep36:
Another option we discussed was patching [net.async](https://github.com/tonsky/net.async) as that is what uses the `InetSocketAddress` object. It hasn't been updated in 8 years, so it'll either get merged in or we could use the fork in our codebase. 

Not sure which is preferable.

jakep36:
This has corrected the issue I have in the flhubee ledger group deployment.

src/fluree/db/ledger/consensus/tcp.clj
```diff
@@ -200,41 +200,57 @@
...
+  "Validates that we are given a server address that we can connect to. Will spin until
+  the address is resolvable to facilitate startup in environments that don't have a DNS
+  entry for a given hostname."
+  [host port]
+  (assert (string? host))
+  (assert (int? port))
+  (loop [retries 0]
+    (let [addr (InetSocketAddress. ^String host ^Integer port)]
+      (if (.isUnresolved addr)
+        (do (Thread/sleep 5000)
```
cap10morgan, line 239: Should we put the sleep after the log so we see it closer to the time it wasn't resolvable?

src/fluree/db/ledger/consensus/tcp.clj
```diff
@@ -200,41 +200,57 @@
...
+            (log/info (str "Remote server address is not resolvable:" host ":" port ". Retrying (" retries ").") )
```
cap10morgan, line 240: Should this be at `warn` level?

cap10morgan (approved):
Looks like there are some changes related to running multiple instances in the same process / REPL, which is fine. Just FYI in case we need to revert anything later, we may not want to revert the PR merge completely.

I do think this is better in the underlying library code, fork or not. But this is fine for now.

I don't think it needs to be configurable.

I asked a couple of questions on specific lines, but overall 👍🏻👍🏻

PR #116, opened 2021-12-13
Only docker build on host platform; run on requested

cap10morgan:
I realized that we were being very inefficient in our Docker builds of ledger b/c the build phase was platform agnostic (it's all Java bytecode, HTML, JS, etc.) but we were building on every requested platform. Adding the `--platform=$BUILDPLATFORM` flag to the builder FROM line fixes this. It runs the build once on the non-emulated builder host's platform (fast & reliable) but then wraps the result in a platform-appropriate openjdk image for the runner (simple & fast even under emulation b/c it isn't doing that much). All we really need is for the `java` binary in the final image to be platform-appropriate, and this still does that. I tested the resulting image on both linux/arm64 and linux/amd64 and they seemed to work just fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM clojure:tools-deps-1.10.3.998-slim-bullseye AS builder
+FROM --platform=$BUILDPLATFORM clojure:openjdk-11-tools-deps-1.10.3.1040-slim-bullseye AS builder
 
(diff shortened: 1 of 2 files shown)


jakep36 (approved):
This looks like an improvement to me. I'd rather run a build and get an error for not setting an ENV var than to accidentally run the build in emulation mode.

PR #120, opened 2022-01-06
FC-1330 - Add /fdb/command endpoint for new-db (mostly)

cap10morgan:
We need an API route that doesn't embed network/ledger in the path for new-db b/c it doesn't exist yet there and will be specified inside the command itself.

Tested this with a (locally modified; separate PR incoming) flurl that signs commands sent to this endpoint. In theory this supports any command, but I tested it with `new-db` as that was the primary (only?) one that didn't already have a ledger to reference.

Part of FC-1330

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -829,12 +829,12 @@
         ledger       (some-> cmd :ledger keyword)
         timeout      (or (:request-timeout headers) 60000)
         opts         {:timeout timeout}
-        _            (log/info "Sending command:" (pr-str cmd))
+        _            (log/debug "Sending command:" (pr-str cmd))
         [headers body] (<?? (action-handler :command system cmd nil ledger
                                             opts))
         request-time (- (System/nanoTime) start)]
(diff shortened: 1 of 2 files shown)


src/fluree/db/peer/http_api.clj
```diff
@@ -814,10 +819,30 @@
...
+  (let [start        (System/nanoTime)
+        cmd          (decode-body body :json)
+        ledger       (some-> cmd :ledger keyword)
+        timeout      (or (:request-timeout headers) 60000)
+        opts         {:timeout timeout}
+        _            (log/info "Sending command:" (pr-str cmd))
+        [headers body] (<?? (action-handler :command system cmd nil ledger
+                                            opts))
+        request-time (- (System/nanoTime) start)]
+    (log/info "Command result:" (pr-str body))
```
dpetran, line 836: Do we want to log this at the info level for every request?

cap10morgan: Nah, probably not. I'll change it to debug.

dpetran (approved):

PR #121, opened 2022-01-07
FC-1338 - Use alts?? to throw exceptions in storage-handler

cap10morgan:
...in http-api. Depends on the addition of util.async/alts?? to db library.

Part of the fix for FC-1338

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -22,7 +22,7 @@
             [fluree.db.peer.websocket :as websocket]
             [ring.util.response :as resp]
             [ring.middleware.cors :as cors]
-            [fluree.db.util.async :refer [<?? <? go-try]]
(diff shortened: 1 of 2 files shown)


src/fluree/db/peer/http_api.clj
```diff
@@ -863,9 +863,9 @@
         ;; final 404 fallback
         (constantly not-found))
 
+      (->> (wrap-errors (:debug-mode? system)))
```
dpetran, line 866: Just for my education - what does moving this up the chain change?

cap10morgan: It seemed more appropriate because the first middleware wrapper's handler will invoke all of the others when it calls `(handler request)` whereas it will have already returned if it's lower down in the list. So this ordering should catch errors even if they are thrown in middleware when it does `(try (handler request) (catch ...))`.

dpetran (approved):

PR #125, opened 2022-01-18
Pass properties file to server startup when running command

dpetran:
With the new :reindex command, we need a valid properties file to pass to the server so
that we can start it up and reindex.


zonotope (approved):
⛏️

jakep36 (approved):
👍

PR #129, opened 2022-01-27
FC-1342 - Feature: multiple root ledger owners

cap10morgan:
This is the ledger side of multiple root owner support for new ledgers. We already have the ability to add owners to a ledger after it's been created, but this allows specifying multiple owners at creation time.

I take advantage of this to allow the http-sig signer to own a new ledger they create, which wasn't supported previously.

db PR: https://github.com/fluree/db/pull/146

JIRA ticket: FC-1342


zonotope (approved):
🎋

PR #130, opened 2022-01-28
FC-1355 - Feature: trust http sigs for transactions

cap10morgan:
ledger counterpart to https://github.com/fluree/db/pull/147

JIRA ticket: FC-1355

Builds on #129 so review that one first :)

I implemented this by adding a new `:signed` flake type that holds the string we signed (to get the `:sig` flake) when that string is something other than the command in the `:tx` flake (which it previously always was before this PR).

### DIFF
--- a/src/fluree/db/ledger/consensus/update_state.clj
+++ b/src/fluree/db/ledger/consensus/update_state.clj
@@ -72,31 +72,43 @@
                              current-val)))]
     (= proposed-v (get-in new-state ks))))
 
+(defn- extract-flake-object
+  "Returns flake object (.-o flake) from block-map of the given type whose
+  subject matches (:t tx-data). Returns nil if none is found."
+  [{:keys [flakes] :as _block-map} {:keys [t] :as _tx-data} type]
+  (some #(let [^Flake flake %]
+           (when (and (= type (.-p flake))
+                      (= t (.-s flake)))
+             (.-o flake)))
+        flakes))
+
 (defn register-new-dbs
   "Register new dbs. Part of state-machine. Updates state-atom, and publishes out :new-db on event-bus"
   [txns state-atom block-map]
-  (let [init-db-status (->> txns
-                            (filter #(and (= :new-db (:type (val %)))
-                                          (= 200 (:status (val %)))))
-                            (map (fn [[_ tx-data]]
-                                   (let [t             (:t tx-data)
-                                         orig-cmd      (some #(when (and (= constants/$_tx:tx (.-p ^Flake %))
-                                                                         (= t (.-s ^Flake %)))
-                                                                (.-o ^Flake %))
-                                                             (:flakes block-map))
-                                         orig-sig      (some #(when (and (= constants/$_tx:sig (.-p ^Flake %))
-                                                                         (= t (.-s ^Flake %)))
-                                                                (.-o ^Flake %))
-                                                             (:flakes block-map))
-                                         orig-cmd-data (when orig-cmd (json/parse orig-cmd))
-                                         {:keys [db fork forkBlock]} orig-cmd-data
-                                         [network dbid] (when orig-cmd (if (sequential? db) db (str/split db #"/")))]
-                                     [network dbid (util/without-nils
-                                                     {:status    :initialize
-                                                      :command   {:cmd orig-cmd
-                                                                  :sig orig-sig}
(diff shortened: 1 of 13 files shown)


src/fluree/db/ledger/consensus/update_state.clj
```diff
@@ -72,31 +72,43 @@
                              current-val)))]
     (= proposed-v (get-in new-state ks))))
 
+(defn- extract-flake-object
+  "Returns flake object (.-o flake) from block-map of the given type whose
+  subject matches (:t tx-data). Returns nil if none is found."
+  [{:keys [flakes] :as _block-map} {:keys [t] :as _tx-data} type]
+  (some #(let [^Flake flake %]
```
zonotope, line 79: you can use `flake/s`, `flake/p`, and `flake/o` below to avoid needing to carry around these type hints.

src/fluree/db/ledger/consensus/update_state.clj
```diff
@@ -72,31 +72,43 @@
...
+             (filter #(and (= :new-db (:type (val %)))
+                           (= 200 (:status (val %)))))
+             (map (fn [[_ tx-data]]
+                    (let [efo           (partial extract-flake-object block-map
+                                                 tx-data)
+                          orig-cmd      (efo constants/$_tx:tx)
+                          orig-sig      (efo constants/$_tx:sig)
+                          orig-signed   (efo constants/$_tx:signed)
+                          orig-cmd-data (when orig-cmd (json/parse orig-cmd))
+                          {:keys [db fork forkBlock]} orig-cmd-data
```
zonotope, line 99: you can combine lines 98 and 99

src/fluree/db/peer/http_api.clj
```diff
@@ -206,19 +206,31 @@
...
-          _           (when-not (sequential? param)
-                        (throw (ex-info (str "A transaction submitted to the 'transact' endpoint must be a list/vector/array.")
-                                        {:status 400 :error :db/invalid-transaction})))
-          txid-only?  (some-> (get opts "txid-only") str/lower-case (= "true"))
-          auth-id     (:auth auth-map)
-          result      (<? (fdb/transact-async conn ledger param {:auth        auth-id
-                                                                 :private-key private-key
-                                                                 :txid-only   txid-only?
-                                                                 :timeout     timeout}))]
+    (let [conn          (:conn system)
```
zonotope, line 209: you can deref this in the argument vector

src/fluree/db/peer/http_api.clj
```diff
@@ -206,19 +206,31 @@
...
+                              ;; original request
+                              (assoc :auth (:authority auth-map))))
+          _             (when-not (sequential? param)
+                          (throw (ex-info (str "A transaction submitted to the "
+                                               "'transact' endpoint must be a "
+                                               "list/vector/array.")
+                                          {:status 400
+                                           :error  :db/invalid-transaction})))
+          txid-only?    (some-> opts (get "txid-only") str/lower-case (= "true"))
+          auth-id       (:auth auth-map)
```
zonotope, line 226: this is probably outside the scope of this pr, but all the `auth` and `authority` names flying around makes this a bit hard to follow.

zonotope (approved):

src/fluree/db/peer/http_api.clj
cap10morgan: I agree. Not sure what better names are, and I have `s/db/ledger/g` higher on my priority list before even digging into this one (someday).

cap10morgan:
Talked this over with @bplatz. I'm going to verify that #1 below is true or implement #2 if not:

1. The value we're already storing in the `tx/tx` flake, when hashed w/ SHA-256, matches the digest value in the `tx/signed` flake.
2. Instead of the internal command, we store the original request body in the `tx/tx` flake so that hashing it gives you the digest value in the `tx/signed` flake.

cap10morgan:
@zonotope Sorry, there's a few more commits after your approval to ensure that you can cryptographically recreate the data provenance from what's stored in the flakes. The main change being in  [c8c318a](https://github.com/fluree/ledger/pull/130/commits/c8c318ab8fa75d3d6750047cebfbc2d19f521ff5) to ensure that the value whose SHA-256 hash is in the signing string's `digest` element is stored in the `tx/tx` flake. Mind taking another look?

PR #131, opened 2022-02-01
Fix/txn error not propagated

ldw1007:
Errors from predicate specs (e.g., unique) were being swallowed and not propagated up to the API layer.  The results 

1) a timeout was returned through the API
2) the transaction was being re-queued (multiple times) for processing. This blocked newer transactions from being processed.  

Since go/try inside a function didn't fix the issue; added a try/block to catch the ExceptionInfo and return it for processing.

Also added a unit test for this scenario.

### DIFF
--- a/src/fluree/db/ledger/transact/validation.clj
+++ b/src/fluree/db/ledger/transact/validation.clj
@@ -566,8 +566,10 @@
               result-ch (async/chan parallelism)
               af        (fn [f res-chan]
                           (async/go
-                            (let [fn-result (as-> (f tx-state*) res
-                                                  (if (channel? res) (async/<! res) res))]
+                            (let [fn-result (try
+                                              (as-> (f tx-state*) res
+                                                    (if (channel? res) (async/<! res) res))
+                                              (catch Exception e e))]
                               (when-not (nil? fn-result)
                                 (async/put! res-chan fn-result))
                               (async/close! res-chan))))]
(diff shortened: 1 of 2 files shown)


src/fluree/db/ledger/transact/validation.clj
```diff
@@ -566,8 +566,10 @@
               result-ch (async/chan parallelism)
               af        (fn [f res-chan]
                           (async/go
-                            (let [fn-result (as-> (f tx-state*) res
-                                                  (if (channel? res) (async/<! res) res))]
+                            (let [fn-result (try
+                                              (as-> (f tx-state*) res
+                                                    (if (channel? res) (async/<! res) res))
+                                              (catch Exception e e))]
```
cap10morgan, line 572: This is what `go-try` and `<?` are intended to help with. Would that work here? Something like (starting on line 568):

```clojure
(go-try
  (let [fn-result (f tx-state*)]
    (if (channel? fn-result)
      (<? fn-result)
      fn-result)))
```

test/fluree/db/ledger/docs/transact/transactions.clj
```diff
@@ -85,6 +85,27 @@
...
+              (= 400)))
+      (is (-> failure-resp
+              ex-data
+              :error
+              (= :db/invalid-tx)))
+      (is (-> failure-resp
+              ex-message
+              (str/starts-with? "Unique predicate person/handle with value: ajohnson matched an existing subject"))))))
+
+
```
cap10morgan, line 108: 👍🏻

cap10morgan (approved):

PR #132, opened 2022-02-02
Enforce restrictCollection property [FC-1345]

ldw1007:
@bplatz There are a couple of places that we can check whether or not a `ref` object is in the designated`restrictCollection`.  I added the `restrictCollection` check to the point where we validate that the `ref` id exists.  We could push the check later in the process, perhaps when we validate predicate/collection specs.

---
At some point, we lost the validation that a 'ref' id is actually contained in the `restrictCollection`.  This change applies the check when the `ref` is verified to exist.

Added a unit test as well.

### DIFF
--- a/src/fluree/db/ledger/transact/core.clj
+++ b/src/fluree/db/ledger/transact/core.clj
@@ -71,6 +71,22 @@
             resolved))))))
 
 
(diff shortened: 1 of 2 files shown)


bplatz (approved):
@ldw1007 This looks great!

PR #133, opened 2022-02-07
FC-1364: Orphaned Index Nodes

zonotope:
This patch refactors the indexing process to separate reading/writing indexes to/from storage from the actual algorithm of integrating novelty flakes into those nodes and rebalancing them as necessary. 

The indexing algorithm itself is now specified as a [transducer](https://github.com/fluree/ledger/pull/133/files#diff-ac9a23749e407a3ebb3af5cea52bddad5af09dea3256c5fe1c5dc973448e9471R191-R250) that transform a stream of index nodes in depth-first order into another stream of nodes, in the same order, with the novelty incorporated and nodes rebalanced as necessary. This transducer doesn't care about where the nodes came from or where they are going, so it works equally on a core.async channel, lazy sequence, or vector. This makes it possible to write tests for the indexing process isolated from any storage or i/o concerns. 

I made this refactor because of the bug described in [FC-1364](https://fluree.atlassian.net/browse/FC-1364). After reindexing databases, some nodes could end up missing from the result because the indexing process failed to consider whether or not a branch/leaf pair were both `leftmost?` when deciding if the leaf was a descendant of the branch. The fix for this bug was simple once it was identified, but that it appeared at all showed the folly of punting on indexing tests during the tspo refactor.

There could definitely be more tests here, but the framework is now in place that will allow us to add them over time.

### DIFF
--- a/src/fluree/db/ledger/indexing.clj
+++ b/src/fluree/db/ledger/indexing.clj
@@ -87,17 +87,6 @@
               (resolve-if-novel conn child t novelty remove-preds error-ch)))
        (async/map vector)))
 
-(defn descendant?
-  [{:keys [rhs leftmost?], cmp :comparator, first-flake :first, :as branch}
-   {node-first :first, node-rhs :rhs, :as node}]
-  (if-not (index/branch? branch)
-    false
-    (and (or leftmost?
-             (not (pos? (cmp first-flake node-first))))
-         (or (nil? rhs)
-             (and (not (nil? node-rhs))
-                  (not (pos? (cmp node-rhs rhs))))))))
-
 (defn filter-predicates
   [preds & flake-sets]
   (if (seq preds)
@@ -176,9 +165,10 @@
          (map (fn [kids]
                 (update-branch branch idx kids)))
          (fn [[maybe-leftmost & not-leftmost]]
-           (into [maybe-leftmost] (map (fn [non-left-node]
-                                         (assoc non-left-node
-                                                :leftmost? false)))
+           (into [maybe-leftmost]
+                 (map (fn [non-left-node]
+                        (assoc non-left-node
+                               :leftmost? false)))
                  not-leftmost)))))
 
 (defn integrate-novelty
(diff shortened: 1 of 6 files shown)


src/fluree/db/ledger/indexing.clj
```diff
@@ -275,15 +275,15 @@
       children)))
 
 (defn descendant?
```
dpetran: Can we add a docstring here to make clear which is the potential ancestor and which the potential descendant? Just to make it easier to use in the future when we've forgotten precisely which way the arguments were arranged : )

src/fluree/db/ledger/indexing.clj
```diff
@@ -159,75 +168,86 @@
...
-         (if (index/leaf? nxt)
-           (let [new-flakes (index/novelty-subrange nxt t novelty)
-                 to-remove  (if (seq remove-preds)
-                              (->> new-flakes
-                                   (concat (:flakes nxt))
-                                   (filter (fn [f]
-                                             (contains? remove-preds
-                                                        (flake/p f)))))
-                              [])]
+        ;;   1. Incorporate each successive node with it's corresponding novelty
```
dpetran: Want to applaud the docs here, they are very helpful. Got an extraneous apostrophe, though :p

src/fluree/db/ledger/indexing.clj
```diff
@@ -197,7 +198,7 @@
          (xf))
 
         ;; Iteration:
-        ;;   1. Incorporate each successive node with it's corresponding novelty
+        ;;   1. Incorporate each successive node with its corresponding novelty
```
dpetran, line 208: Ope, you fixed it. That's what I get for going through in commit order :1st_place_medal:

src/fluree/db/ledger/indexing.clj
dpetran: And I see you moved this to the `db` lib, I'll move this comment there.

src/fluree/db/ledger/indexing.clj
```diff
@@ -107,216 +87,230 @@
...
-                    out)))
-    out))
-
-(defn tally
-  [idx index-ch stat-ch]
+         (partition-all target-count)
+         (map (fn [kids]
+                (update-branch branch idx t kids)))
+         (fn [[maybe-leftmost & not-leftmost]]
+           (into [maybe-leftmost]
```
dpetran, line 185: I never knew about this arity for `into`! I think you may need to double-wrap this `fn` in parens in order for it to play nicely with the `->>`, though, or move it to a `defn`.

dpetran (approved):

PR #135, opened 2022-02-16
FC-1367 - Feature: http request id header

cap10morgan:
If there is a request header `X-Fdb-Request-Id`, save its value and return it in the response in the HTTP API. For Nexus request tracing.

Implements FC-1367


cap10morgan:
@fluree/core This is ready for review now. Thanks!

dpetran (approved):

PR #136, opened 2022-02-17
FC-1335, FC-1336, FC-1339, FC-1340 Updated Purge Flakes Logic  

ldw1007:
Dependent on db pull request: https://github.com/fluree/db/pull/154

----
Multiple updates:
1. Block files are deleted versus renamed [FC-1335]
2. Full-text indexes are cleaned up [FC-1340]
3. Clean-up dependency refs that are designated as a `component`(similar to a cascade delete) [FC-1336]
4. Clean-up subject references to a "parent" [FC-1339]
5. Added unit tests [FC-1337]

[FC-1335]: https://fluree.atlassian.net/browse/FC-1335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FC-1340]: https://fluree.atlassian.net/browse/FC-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FC-1336]: https://fluree.atlassian.net/browse/FC-1336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FC-1339]: https://fluree.atlassian.net/browse/FC-1339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FC-1337]: https://fluree.atlassian.net/browse/FC-1337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fluree/db/ledger/mutable.clj
```diff
@@ -58,15 +60,38 @@
...
                 from-t    (if (and block-start (not= 1 block-start))
-                            (dec (:t (<? (time-travel/as-of-block db (dec block-start))))) -1)
+                            (dec (:t (<? (time-travel/as-of-block db (dec block-start)))))
+                            -1)
                 to-t      (if block-end
-                            (:t (<? (time-travel/as-of-block db block-end))) (:t db))
+                            (:t (<? (time-travel/as-of-block db block-end)))
+                            (:t db))
                 flakes    (<? (query-range/time-range db idx = pattern {:from-t from-t :to-t to-t}))
+                _ (log/info "identify-hide-blocks-flakes " {:flakes flakes})
```
dpetran, line 94: Since users run at INFO level logging by default, we should either clean this up so what it's indicating is more clear, or change it to DEBUG level.

src/fluree/db/ledger/mutable.clj
```diff
@@ -6,7 +6,9 @@
             [fluree.db.storage.core :as storage]
             [fluree.db.constants :as const]
             [fluree.db.util.log :as log]
-            [fluree.db.ledger.reindex :as reindex])
+            [fluree.db.ledger.reindex :as reindex]
+            [fluree.db.util.core :as util]
+            [fluree.db.flake :as flake])
   (:import (fluree.db.flake Flake)))
```
dpetran, line 12: I know this isn't something you added here, but we no longer need the `Flake` type hint, we've got `s` `p` `o` `t` etc helpers in the `flake` namespace that do the hinting for us.

src/fluree/db/ledger/mutable.clj
```diff
@@ -42,10 +44,10 @@
 
 (defn purge-block
   [conn nw ledger block flakes]
-  (go-try (let [conn-rename   (:storage-rename conn)
+  (go-try (let [conn-delete   (:storage-delete conn)
                 block-key     (storage/ledger-block-key nw ledger block)
                 current-block (<? (storage/read-block conn nw ledger block))
-                _             (<?? (conn-rename block-key (str block-key "-delete")))
+                _             (<?? (conn-delete block-key))
```
dpetran, line 50: Is it okay to use a blocking take inside a go block? The docstring for `<!!` says "Not intended for use in direct or transitive calls from (go ...) blocks", but I'm not totally familiar with this case.

src/fluree/db/ledger/mutable.clj
```diff
@@ -58,15 +60,38 @@
...
+
+(defn- create-block-map
+  "Creates a block-map from a vector of flakes"
+  [db flakes]
+  (go-try
+    (loop [[^Flake flake & r] flakes
+           t-map     {}
+           block-map {}]
+      (if flake
+        (if-let [block (get t-map (.-t flake))]
```
dpetran, line 72: Instead of calling the `-t` method directly, can we use the `flake/t` helper function so we don't need the type hint?

src/fluree/db/ledger/mutable.clj
ldw1007: I believe that the original contributor's intent was for the rename/delete operation to be blocking.  The purge operation should fail if even one block could not be purged.

We probably need to test a bit more to verify that the blocking actually happens.

src/fluree/db/ledger/mutable.clj
ldw1007: Changed to debug

src/fluree/db/ledger/mutable.clj
ldw1007: replaced all ^Flake references (that I found) with the appropriate `flake/` helper function.

src/fluree/db/ledger/mutable.clj
dpetran: I tried a basic example at the repl and it seemed to work, but I don't know what the consequences are or why that warning is there.

dpetran: https://fluree.atlassian.net/browse/FC-1369

PR #138, opened 2022-03-02
Tests: signing history & block queries

cap10morgan:
Once I got the tests working correctly they passed, so it seems like /history and /block work OK w/ http-sig requests


zonotope (approved):
📚

PR #139, opened 2022-03-05
preserve id under the old-id for nodes updated during indexing

zonotope:
This fixes a bug where new ledgers can't be created (and old ledgers can't be indexed) I introduced by trying to DRY up tree walking code without accounting for downstream code that relies on the previous id from index nodes to be preserved under the `old-id` key. The indexing process itself succeeds, but no garbage could be serialized because the garbage keys were not preserved. 

This patch adds back the old-id key for updated nodes, allowing them to be serialized.

### DIFF
--- a/src/fluree/db/ledger/indexing.clj
+++ b/src/fluree/db/ledger/indexing.clj
@@ -210,6 +210,10 @@
              (xf result*))))))))
 
 (defn preserve-id
+  "Stores the original id of a node under the `::old-id` key if the `node` was
+  resolved, leaving unresolved nodes unchanged. Useful for keeping track of the
+  original id for modified nodes during the indexing process for garbage
+  collection purposes"
   [{:keys [id] :as node}]
   (cond-> node
     (index/resolved? node) (assoc ::old-id id)))
--- a/src/fluree/db/ledger/indexing.clj
+++ b/src/fluree/db/ledger/indexing.clj
@@ -209,6 +209,11 @@
                     (unreduced (xf result* node)))
              (xf result*))))))))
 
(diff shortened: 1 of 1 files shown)


dpetran (approved):

dpetran:
This looks good to me, is there a test we could have that would have caught this? We don't necessarily have to add it here, but I'd like to slowly build up our regression test suite.

zonotope:
> This looks good to me, is there a test we could have that would have caught this? We don't necessarily have to add it here, but I'd like to slowly build up our regression test suite.

I was thinking about how best to add a test. The difficulty is that I think it requires a full end-to-end integration test that actually serializes the index (and garbage files) to disk. I'll see if I can do that today, but I'll also merge this now without it just so the broken main doesn't block anyone.

cap10morgan:
> > This looks good to me, is there a test we could have that would have caught this? We don't necessarily have to add it here, but I'd like to slowly build up our regression test suite.
> 
> 
> 
> I was thinking about how best to add a test. The difficulty is that I think it requires a full end-to-end integration test that actually serializes the index (and garbage files) to disk. I'll see if I can do that today, but I'll also merge this now without it just so the broken main doesn't block anyone. 
> 
> 

FYI I have a working integration test locally that I'll PR today. I wrote it on the plane on Friday.

It just tries to create, transact against, and query a ledger but using file storage instead of memory storage like the rest of the test suite.

It fails with this issue and I'll try it again with this fix merged to make sure it now passes.

cap10morgan:
The aforementioned test: https://github.com/fluree/ledger/pull/140

PR #140, opened 2022-03-08
Test file storage ledger create & interaction & fix race condition in tests

cap10morgan:
This does a couple of things:

1. It adds an integration test that creates and uses a ledger w/ file storage instead of the memory storage the rest of the tests use. There was a recent bug that only appeared with file storage, and it isn't the first one of its kind. Now we'll have a test that can catch some of those.
2. It fixes a somewhat long-standing race condition in the test suite where it starts trying to create ledgers right after spinning up a test system. Every once in awhile it sent a `:new-db` command before the default private key was added to the group's local-state, and then the command would fail. The test suite now waits for that private key to appear before running the tests. This seems unlikely to come up in production, though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,12 @@ target/fluree-ledger.jar: resources/adminUI $(SOURCES) $(RESOURCES)
 
 jar: target/fluree-ledger.jar
 
+unit-test:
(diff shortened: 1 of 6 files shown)


dpetran (approved):
:1st_place_medal:

PR #144, opened 2022-03-14
FC-1360 Context smartfunctions - bootstrapping and tests - updated on main

bplatz:
Companion to corresponding fluree/db release incorporating context and associate smartfunctions. See: https://github.com/fluree/db/pull/161

Includes new test for feature in 'invoicing' example.


test/fluree/db/ledger/general/invoice_tests.clj
```diff
@@ -97,7 +97,74 @@
...
+              {"_user/username" "scott", :_id 87960930223082}
+              {"_user/username" "freddie", :_id 87960930223081}]))
+      ;; scott can see only _user in the same org
+      (is (= scott-q-res
+             [{:_id 87960930223084, "_user/username" "fannie"}
+              {:_id 87960930223082, "_user/username" "scott"}]))
+      ;; antonio can only see himself, no other users are in same org
+      (is (= antonio-q-res
+             [{"_user/username" "antonio", :_id 87960930223083}])))))
+
```
cap10morgan, line 163: Love the addition of tests!

One thing to consider is that when we migrate this test ns to be more idiomatic, each `deftest` will need to be runnable in isolation w/ its own randomly-named ledger. In here the second `deftest` relies on running after the first one & in the same ledger. It's fine if you want to leave changing that until we get to migrating this ns. Or you could change that now or even fully migrate the test ns to idiomatic if you want to / have time (I [documented the process here](https://github.com/fluree/ledger/blob/main/test/CONVERT_TO_IDIOMATIC.md)).

I think whichever option you choose is fine, just wanted to point it out.

cap10morgan (approved):
Looks good! Yay tests!

PR #150, opened 2022-04-06
FC-1382: Delete stale files after reindexing

zonotope:
This patch identifies all index nodes descended from the current index root, and deletes every other file directly from that Raft group and storage.

### DIFF
--- a/src/fluree/db/ledger/reindex.clj
+++ b/src/fluree/db/ledger/reindex.clj
@@ -1,8 +1,10 @@
 (ns fluree.db.ledger.reindex
   (:require [fluree.db.util.log :as log]
             [fluree.db.storage.core :as storage]
+            [fluree.db.ledger.storage :as ledger-storage]
             [fluree.db.dbproto :as dbproto]
             [fluree.db.flake :as flake]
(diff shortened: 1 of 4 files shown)


src/fluree/db/ledger/reindex.clj
```diff
@@ -172,19 +209,23 @@
...
-                         network           (:network indexed-db)
-                         dbid              (:dbid indexed-db)
-                         index-point       (get-in indexed-db [:stats :indexed])
-                         state-atom        (-> conn :group :state-atom)
-                         submission-server (get-in @state-atom [:_work :networks network])]
-                     (<? (txproto/write-index-point-async group network dbid index-point submission-server {}))
-                     indexed-db)                            ;; final index if any novelty
-                   db))
+                 (let [final-db (if (> (get-in db [:novelty :size]) 0)
+                                  (let [indexed-db        (async/<! (indexing/refresh db {:status  status
```
cap10morgan, line 213: Can this return an error? If so, should we switch it to `<?`?

cap10morgan (approved):
One question but looks good overall!

PR #151, opened 2022-04-07
FC-1385: Retry all filestore reads 3 times

zonotope:


dpetran (approved):
This looks good - does address some sort of filesystem race condition issue, or is this just preventative? Should we consider adding retries in other locations?

zonotope:
> This looks good - does address some sort of filesystem race condition issue, or is this just preventative? Should we consider adding retries in other locations?

I believe it's a filesystem race condition issue, but it only happens transiently, and for very large databases so it's hard to pin down what's happening. We're currently doing the retry manually when this occurs, but that means downtime. 

I don't think this is the ultimate fix, but it at least allows the database to move forward when these transient issues occur. We've only seen this for file-backed storage, so I think it's a filesystem/disk issue, so I tried to keep the retry logic as focused as possible to just that code path.

PR #156, opened 2022-04-27
Ensure that :txid-only defaults to false in /command HTTP API endpoint

cap10morgan:
The old code defaulted to true when the param was omitted, unlike everywhere else where it defaults to false.

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -200,18 +200,6 @@
         true))))
 
 
-(defn- default-to-false
-  "Returns the value of key k in map m with default value `false`. Meaning that
-  if the value is anything but `true`, this will return `false`. Iff the value
-  is `true`, it will return `true`.
-
-  The difference between this and (get m k false) is that will return `true`
-  when the value is truthy (e.g. a string) whereas this treats anything other
-  than `true` as `false`."
-  [m k]
-  (-> m (get k) true?))
-
-
 (defmulti action-handler (fn [action _ _ _ _ _] action))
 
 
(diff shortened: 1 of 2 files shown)


cap10morgan:
Reminder to self: Merge maintenance/v1 into main after this lands.

src/fluree/db/peer/http_api.clj
```diff
@@ -275,7 +287,8 @@
           result (cond
                    (and (:cmd param) (:sig param))
                    (let [persist-resp (<? (fdb/submit-command-async conn param))
-                         result       (if (and (string? persist-resp) (-> param :txid-only false?))
+                         result       (if (and (string? persist-resp)
+                                               (not (default-to-false param :txid-only)))
```
bplatz, line 291: Maybe instead of a new fn?:
```
(if (true? (:txid-only param))
   persist-resp
   (<? (fdb/monitor-tx-async conn ledger persist-resp timeout)))
```

bplatz (approved):
One suggestion at your option, but otherwise LGTM.

src/fluree/db/peer/http_api.clj
cap10morgan: Yeah that's way better. Do you think it's still important to check for the string value in `persist-resp`?

bplatz: I think so. Missed that in my response. If it isn't a string, I think it means there is an error, and with <? it should check/throw but maybe some other condition case exists that makes sense to leave the check in there for now bing this is a quick fix.

PR #160, opened 2022-05-20
Bug/reify db failure

zonotope:
This patch is mainly picking up the changes in https://github.com/fluree/db/pull/174, as well as some cleanup and fixing the tests to be a bit more deterministic to explicitly wait on a db with a specified block.


Dockerfile
```diff
@@ -20,7 +20,7 @@ COPY . ./
 RUN make uberjar
 RUN make stage-release
 
-FROM openjdk:17-slim-bullseye AS runner
+FROM eclipse-temurin:17-jre-focal AS runner
```
dpetran, line 23: What's different about this image?

dpetran: Ah, I see this came in from the merge, nvm

dpetran:
Oops, I just saw this was still in-progress, I'll leave you alone until it's done!

PR #161, opened 2022-06-04
FC-1409 add test for two-tuple where binding, update db dependency

bplatz:


test/fluree/db/ledger/docs/query/advanced_query.clj
```diff
@@ -121,6 +121,24 @@
...
+    (let [query      {:select ["?e" "?maxNum"]
+                      :where  [["?e" "person/favNums" "?favNums"]
+                               ["?maxNum" "#(max ?favNums)"]]}
+          db         (basic/get-db test/ledger-chat)
+          res        (async/<!! (fdb/query-async db query))
+          maxNumVals (->> res
+                          (map second)
+                          (into #{}))]
+
+      ;; all two-tuple results should have the identical ?maxNum
```
zonotope, line 135: I think these comments would be more useful in the optional `msg` argument of the [`is` test macro](https://clojuredocs.org/clojure.test/is). That way if/when this test fails, we'll get a printed error message in the console/repl:

```clojure
(is (= 1 (count maxNumVals))
    "all two-tuple results should have the identical ?maxNum")
```

test/fluree/db/ledger/docs/query/advanced_query.clj
```diff
@@ -121,6 +121,24 @@
...
+      (is (= 1 (count maxNumVals)))
+
+      ;; sample data set has max favNum as 1950
```
zonotope, line 138: same here as above 👆🏾

zonotope (approved):
Looks good to me besides minor documentation suggestions.

PR #164, opened 2022-06-09
Fix FC-1414 block not found

dpetran:
This also includes a fix for the broken new-ledger command - it was checking for a keyword and we were supplying a string.

This is a fix for `maintenance/v1`, we may also need to apply these changes to main, but I haven't checked yet.

### DIFF
--- a/src/fluree/db/peer/http_api.clj
+++ b/src/fluree/db/peer/http_api.clj
@@ -779,10 +779,10 @@
                                            (throw (ex-info (str "If using a closed api, a storage request must be returned as json.")
                                                            {:status 401
                                                             :error  :db/invalid-transaction})))
-                        {:keys [network ledger type key]} params
-                        ledger           (keyword network ledger)
-                        _                (when-not (and network ledger type)
-                                           (throw (ex-info (str "Incomplete request. At least a network, db and type are required. Provided network: " network " ledger: " ledger " type: " type " key: " key)
+                        {network :network dbid :ledger type :type key :key} params
+                        ledger        (keyword network ledger)
+                        _                (when-not (and network dbid type)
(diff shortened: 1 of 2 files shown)


src/fluree/db/peer/http_api.clj
```diff
@@ -779,10 +779,10 @@
                                            (throw (ex-info (str "If using a closed api, a storage request must be returned as json.")
                                                            {:status 401
                                                             :error  :db/invalid-transaction})))
-                        {:keys [network ledger type key]} params
-                        ledger           (keyword network ledger)
-                        _                (when-not (and network ledger type)
-                                           (throw (ex-info (str "Incomplete request. At least a network, db and type are required. Provided network: " network " ledger: " ledger " type: " type " key: " key)
+                        {network :network dbid :ledger type :type key :key} params
+                        ledger        (keyword network ledger)
```
bplatz, line 783: Should this be `(keyword network dbid)` ?

bplatz (approved):
Looks good, only main concern is making sure the (keyword ...) fn I commented on is correct. If so, good here!

src/fluree/db/peer/http_api.clj
dpetran: Good catch, thanks.

src/fluree/db/peer/http_api.clj
```diff
@@ -811,9 +811,11 @@
                                                             (throw (ex-info (str "Auth id for request does not exist in the database: " jwt-auth)
                                                                             {:status 403 :error :db/invalid-auth})))]
                                              jwt-auth))
-                        formatted-key    (cond-> (str network "_" ledger "_" type)
-                                                 key (str "_" key))
+                        formatted-key    (cond-> (str network "_" dbid "_" type)
+                                           key (str "_" key))
```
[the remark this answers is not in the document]
dpetran: I've added it there and will change it here once db gets published. There's now a todo to remind us.

PR #165, opened 2022-06-15
restore main docker tag

cap10morgan:
@bplatz asked if we could bring this back, but only for the main branch. So hopefully this won't clutter up our Docker tags too badly.

I also implemented a post-1.0.0 TODO I noticed in here.


zonotope (approved):
🏷️

PR #167, opened 2022-06-29
FC-1416: Fix/full text exception

zonotope:
The actual bug fix is [here](https://github.com/fluree/ledger/compare/fix/full-text-exception?expand=1#diff-0ccab2b33b9dd82422bb1d3cc78e9806e72b54270d68c07c127aee7dd4fc6951R40-R50), that extracts the actual full text predicate values from the subject attributes of the updated flakes. The rest of these changes are to give access to the test/example data in development while still maintaining the right logging config both in test and dev by adding a dev-resources directory to contain the test/example data and making that accessible to both the dev and test aliases. I also did a bit of cleanup to use the flake component functions to cut down on all the type hints. 

I tried really hard to add a test that would have caught this bug, but unfortunately the in-memory full text store doesn't hold on to the data after it's closed, and we can't hold it open because it isn't thread safe and can only be opened by one process at a time.


cap10morgan:
Found it: https://github.com/fluree/ledger/blob/main/test/fluree/db/ledger/integration/file_storage_test.clj

zonotope:
@cap10morgan thanks for the pointer! I have a file-backed full text integration test up, and this is ready for review again.

PR #168, opened 2022-06-29
Fix single cardinality tempids

cap10morgan:
As noted in FC-1415, single-cardinality tempids (e.g. commonly seen in ref preds) were not behaving correctly on updates. In analytical query results both the old and new values were being returned.

This appears to have originated with the transaction refactor from PR #40 which was released with v1.0.0-beta11.

The problem seemed to be that no retract flake was being created for the old value in the post-refactor tempid handling logic.

I'm not sure the fix in this PR is the right place / way to fix this, but it does seem to fix it. Let me know if I should do this differently.

### DIFF
--- a/src/fluree/db/ledger/transact/core.clj
+++ b/src/fluree/db/ledger/transact/core.clj
@@ -234,6 +234,7 @@
 (defn resolve-object
   "Resolves object into its final state so can be used for consistent comparisons with existing data."
   [object _id pred-info tx-state]
+  (log/debug "Resolving object:" object "for pred:" pred-info)
   (let [multi? (pred-info :multi)]
     (if (nil? object)
       (async/go :delete)                                    ;; delete any existing object
@@ -255,6 +256,7 @@
   Performs some logic to determine if the new flake should get added at
   all (i.e. if retract-flake is identical to the new flake)."
   [flakes ^Flake new-flake ^Flake retract-flake pred-info tx-state]
+  (log/debug "add-singleton-flake new-flake:" new-flake "retract-flake:" retract-flake)
   (cond
     ;; no retraction flake, always add
     (nil? retract-flake)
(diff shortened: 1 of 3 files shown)


zonotope (commented):
This looks good to me, but I would feel better if @bplatz gave it a glance too to make sure.

src/fluree/db/ledger/transact/core.clj
```diff
@@ -355,16 +358,20 @@
                     (recur acc** r))
 
                   (or tempid? (tempid/TempId? obj*))
-                  (-> acc
-                      (update :_temp-flakes conj (flake/->Flake _id** pid obj* t true nil))
-                      (recur r))
+                  (do
+                    (swap! upserts conj _id**) ; make sure we retract the old value
```
bplatz, line 362: Looks great - so the only condition this issue happened for was upserts?

Meaning, updating a ref on an existing subject that uses an _id value for an existing identity (e.g. `123435456` or two-tuple `[_user/username 'bplatz']`) was working fine?

bplatz (approved):
Great, just one comment to confirm.

src/fluree/db/ledger/transact/core.clj
cap10morgan: No, and this is part of why I'm not sure this is the best / right solution.

Updating refs w/ two-tuples was *not* working; at least in the case where it was a tempid for something being created in the same txn; which is what my test was doing. That's an "upsert" in the sense that because it's not marked `:multi true`, it should update the existing value, but instead it was being added to the predicate's object as if it were `:multi true`. But none of the preds are marked `:upsert true`, so I wasn't sure if the semantics were quite right here. Is there a semantic difference between `upsert` and `update` for an existing subject?

I'm working with [these test preds](https://github.com/fluree/ledger/blob/59697cc698def9021f9a65b8c54b923dc9991287/test/fluree/db/ledger/docs/transact/transactions.clj#L39-L52).

I can write some additional tests for existing ids, etc.

cap10morgan: Gaining more insight as I write additional tests:

Referencing a pre-existing subject w/ a two-tuple *was* working before this change (and still does afterward, just to be clear). So far it seems that the only broken case was referencing a subject in the same transaction with a tempid.

But I'm going to write at least one more test for referencing an existing subject by id instead of two-tuple.

cap10morgan: And referencing pre-existing subjects by id worked and works too.

cap10morgan:
Going to go ahead and merge this once it's green as all I added post-review was a couple more tests to convince myself that the bug was limited to tempids only.

PR #172, opened 2022-07-05
Fix missing bootstrap metadata flakes

cap10morgan:
Fixes FC-1420

Not sure what happened here but I (most likely inadvertently) removed these in b2f9d1a800f16af0e51aa94a5063c6658b3516fb. This adds them back and adds a test to check that they're all there.

This only affects `main` / 2.x.


src/fluree/db/ledger/bootstrap.clj
```diff
@@ -27,6 +27,17 @@
        (crypto/sha3-256)))
 
 
+(defn predicate->id-map
+  "Returns a map of predicate namex to final predicate ids, i.e {\"_user/username\" 10}"
```
zonotope, line 31: namex -> names, i think

zonotope (commented):
👍🏾

PR #173, opened 2022-07-05
Eliminate unnecessary sleeps in tests

cap10morgan:
Closes FC-1245

There only ended up being one place where we were sleeping instead of using `:syncTo`, but I eliminated some other unnecessary sleeps in the tests. It's a few seconds faster overall now. :)


dpetran (approved):
Thanks for wrapping the `testing` statements as well! :pray:

cap10morgan:
> Thanks for wrapping the `testing` statements as well! 🙏

I can't not when I see them 🙈

PR #179, opened 2022-07-26
Update old snapshot format w/ :dbs to :ledgers when we read them in from disk

cap10morgan:
In debugging FC-1428 I noticed that the old raft snapshot data looked like `{:networks {"nw-name" {:dbs {"ledger-id" ...}}}}` and the system couldn't see them as a result b/c it's looking for `:ledgers` now. This adds an upgrade fn that (for now) just replaces those `:dbs` keys with `:ledgers` but leaves room for additional upgrades we may want / need to do to these snapshots down the road. It only does this once on startup and subsequent snapshots should have the new format.

The actual fix is 1ddb5930ab7cec41cf0d5d5c2998948f76fa42de, the other commit is just code reformatting.


zonotope (approved):
📸

PR #188, opened 2022-08-23
FC-1429: use bugfix version of db dependency

zonotope:
The real fix is in https://github.com/fluree/db/pull/194. This patch brings those changes in.


dpetran (approved):
:+1:

PR #190, opened 2022-08-29
FC-814: feature/command pool

zonotope:
This patch adds a new `:put-pool` state machine transition that adds entries to a capped command pool in state. The cap is configurable under the `:fdb-pending-tx-limit` property/environment variable and the default is now 10,000, which is probably pretty high, but I don't have a good picture of what a typical workflow is like. The pool is capped for each individual ledger (network and ledger-id pair), but we can change it if capping on the network or globally is more appropriate. If a new transaction comes in when the command pool is full, the system will throw an exception with a 503 status back to the client, and the client controls when/if to retry.

In an attempt to keep pull requests manageable, I've decided to implement FC-814 in phases, and this is phase 1: adding the command pool and throwing an exception when trying to add to a full pool. Phase 2 will implement backpressure where, instead of throwing exceptions, the server directs connected clients to pause when there's a full queue, and detects when the queue has been drained and tells those clients to continue sending messages. Phase 3 will refactor the other state transitions to use pure functions to both update state, and respond to those state updates so the state management code can be simpler, more modular, and, most importantly, testable.

### DIFF
--- a/src/fluree/db/ledger/txgroup/txgroup_proto.clj
+++ b/src/fluree/db/ledger/txgroup/txgroup_proto.clj
@@ -67,10 +67,10 @@
   (async/<!! (kv-dissoc-in-async group ks)))
 
 (defn kv-max-in-async
-  "Writes value to specified key sequence. Will only write is proposed value is
-  greater than the current value. Returns true if value is written, false if it
-  is not written. Returns core async channel that will eventually have a
-  response."
+  "Writes value to specified key sequence. Will only write if the proposed value
+  is greater than the current value. Returns a core async channel that will
+  eventually have a response of true if the value is written, false if it is not
+  written."
   [group ks v]
   (assert (number? v))
   (let [command [:max-in ks v]]
@@ -325,7 +325,7 @@
            (command-queue group network))))
 
 (defn queue-command-async
-  "Enqueues a new command processing"
+  "Enqueues a new command or transaction for processing."
   [group network ledger-id command-id command]
   (let [queue-time  (System/currentTimeMillis)
         size        (-> command :cmd count)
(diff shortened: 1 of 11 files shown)


src/fluree/db/ledger/txgroup/txgroup_proto.clj
```diff
@@ -45,42 +54,47 @@
...
+  "Remove value as key sequence. Returns true if a value was present and removed,
+  false if no value existed."
   [group ks]
   (async/<!! (kv-dissoc-in-async group ks)))
 
 (defn kv-max-in-async
-  "Writes value to specified key sequence. Will only write is proposed value is greater
-  than the current value. Returns true if value is written, false if it is not written.
-  Returns core async channel that will eventually have a response."
+  "Writes value to specified key sequence. Will only write is proposed value is
```
dpetran, line 70: Not yours, but I just spotted a typo here: "will only write is" -> "will only write if".

src/fluree/db/ledger/txgroup/txgroup_proto.clj
```diff
@@ -284,38 +302,40 @@ or this server is not responsible for this ledger, will return false. Else true
...
-   (->> (get-in (-local-state group) [:cmd-queue network])
-        vals
-        (filter #(= ledger-id (:ledger-id %))))))
+   (filter (fn [cmd-entry]
+             (-> cmd-entry :ledger-id (= ledger-id)))
+           (command-queue group network))))
 
 (defn queue-command-async
-  "Writes a new tx to the queue"
+  "Enqueues a new command processing"
```
dpetran, line 328: I'm not quite clear on this docstring.

dpetran (approved):
This looks good :+1:

cap10morgan:
@zonotope Should we merge this?

zonotope:
this is part of a bigger change i'm still working on, but wanted to break into multiple prs. it is mergable, but i'd rather merge everything together.

zonotope:
Closing for now to focus on json-ld

PR #193, opened 2022-09-07
Add orderBy variable query tests

cap10morgan:
Adds tests for queries that weren't working before https://github.com/fluree/db/pull/196


dpetran (approved):
Thanks for the tests!

PR #194, opened 2022-09-12
Fix homebrew installs

cap10morgan:
This combined with https://github.com/fluree/homebrew-flureedb/pull/4 should fix the homebrew issues caused by their recent move to `/opt/homebrew` for Apple Silicon Macs (and/or our overly-stringent assumptions about which directories to search for system-wide config files, depending on your perspective). :)


cap10morgan:
This should be forward-ported to main once merged.

bplatz (approved):

PR #197, opened 2022-09-15
Add a couple tests for filter vars

cap10morgan:
Tests for filter vars in queries. One of them tests for the fix in https://github.com/fluree/db/pull/201.


zonotope (approved):
⚖️

PR #198, opened 2022-09-21
Support :json pred parsing & add tests for it

cap10morgan:
This goes with https://github.com/fluree/db/pull/203 to add a `:parseJSON true` opt for queries w/ predicates of type `:json`. This also adds some tests for `:json` preds in transactions and queries.

### DIFF
--- a/test/fluree/db/ledger/predicate/json_test.clj
+++ b/test/fluree/db/ledger/predicate/json_test.clj
@@ -9,9 +9,9 @@
 (use-fixtures :once test/test-system)
 
 (deftest transact-json-test
-  (let [ledger (test/rand-ledger "pred/json")
-        _ (test/assert-success
-            (test/transact-schema ledger "json-preds"))
+  (let [ledger  (test/rand-ledger "pred/json")
+        _       (test/assert-success
+                  (test/transact-schema ledger "json-preds"))
         api-url (str "http://localhost:" @test/port "/fdb/" ledger "/")]
     (testing "valid JSON succeeds"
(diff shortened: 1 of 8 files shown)


test/fluree/db/ledger/predicate/json_test.clj
```diff
@@ -0,0 +1,109 @@
...
+                                               :body    txn})]
+        (is (= 200 status))
+        (is (some #(-> % (nth 2) (= "{\"foo\":\"bar\"}")) (-> body json/parse :flakes)))))
+    (testing "invalid JSON fails"
+      (let [txn "[{\"_id\": \"_user\", \"_user/username\": \"tester\", \"_user/json\": {\"foo\": \"bar\"}]" ; missing closing }
+            {:keys [status]} @(http/post (str api-url "transact")
+                                         {:headers {"Content-Type" "application/json"}
+                                          :body    txn})]
+        ;; This should really return a 4xx error, but it's a 5xx error as of the date I wrote this test.
+        ;; So I'm just checking that this is in the "error range" of HTTP status codes.
```
dpetran, line 29: Do we have a fix for this planned? It would nice to be able to assert a specific error code for this so that users get a nice response.

cap10morgan: Not that I'm aware of. I'm not sure how much work it would be, and I also wonder how consistent this is already across different endpoints and errors. Either way, it felt like a low priority to fix, but the test looked weird w/o an explanation.

zonotope (approved):

PR #200, opened 2022-10-13
FC-1437: add tests for order-by with limit and offset

zonotope:
Tests for the code added in https://github.com/fluree/db/pull/207

### DIFF
--- a/test/fluree/db/ledger/query/analytical_test.clj
+++ b/test/fluree/db/ledger/query/analytical_test.clj
@@ -266,7 +266,7 @@
               query-limit (assoc query-all :limit limit)
               subject     (async/<!! (fdb/query-async db query-limit))]
 
-          (is (= subject (into [] (take limit) all-results))
+          (is (= (count subject) limit)
               "returns subjects only up to the limit")
 
           (testing "and with offset"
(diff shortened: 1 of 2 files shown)


test/fluree/db/ledger/query/analytical_test.clj
```diff
@@ -247,6 +248,38 @@
...
+          (testing "and with offset"
+            (let [offset       1
+                  query-offset (assoc query-limit :offset offset)
+                  subject      (async/<!! (fdb/query-async db query-offset))]
+
+              (is (= subject (into []
+                                   (comp (drop offset)
+                                         (take limit))
+                                   all-results))
+                  "returns subjects only up to the limit after the offset"))))))))
```
cap10morgan, line 281: I'd have _slightly_ more confidence in these tests if the expected values were either hard-coded values or calculated/compared in a much more primitive way (vs. the logic implementation in the code under test).

E.g. for offest=1, limit=1
```clojure
(is (= 1 (count subject)))
(is (= (first subject) (second all-results)))
```

Otherwise they could be wrong for the same reason(s) and agree with each other and pass.

Not a huge deal here, so I'll leave it up to you.

cap10morgan (approved):
One comment on the tests I'll leave entirely to your judgment, but certainly not a blocker.

test/fluree/db/ledger/query/analytical_test.clj
zonotope: yeah, i definitely had some hesitance about using exactly the same algorithm too. i think i got too clever in the name of test robustness. i'll change it.

PR #201, opened 2022-11-01
Use new flake/create fn in bootstrap

cap10morgan:
`flake/new-flake` no longer exists, so this uses `flake/create` instead. However, I'm sure some of the datatypes (new `dt` field) are wrong and the test suite ledgers still never initialize even after applying this fix. I'm not sure if that's because this is wrong or because of other issues, though.

### DIFF
--- a/src/fluree/db/ledger/bootstrap.clj
+++ b/src/fluree/db/ledger/bootstrap.clj
@@ -143,47 +143,47 @@
     ;                  {:status 500 :error :db/unexpected-error})))
     (let [base-flakes
           [;; add a true predicate function
-           (flake/new-flake true-fn-sid (get pred->id "_fn/name") "true" t true)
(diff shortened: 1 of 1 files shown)


cap10morgan:
Think I found the issue here for why new ledgers aren't initializing: The value of `const/$_block:hash` was recently changed to `51` which is also the value of `const/$_user:auth` so the block hash flake was getting mistaken for an auth flake. Might be good to add a test that all values in that ns are unique (or something close to that)?

bplatz:
For json-ld, all DB bootstrapping, indexing, and probably a couple other things I'm forgetting are now handled directly by functions in the db library. Historically ledger did this, but now ledger's role is more to facilitate consensus and to expose http APIs. Pretty much any database/ledger related, including indexing, should now be entirely contained in fluree/db.

Until we make those adjustment in ledger's code to defer these operations fully to db, I'm guessing you'll find some inconsistencies like this.

cap10morgan:
@bplatz 10-4. I'll close this PR and approach it from that angle. Thanks!

PR #205, opened 2023-03-31
Feature: -Xmx for commands

cap10morgan:
Modifies the startup script to allow setting -Xmx (etc.) for commands too. `:reindex` sometimes needs this on large datasets, for example.

Snuck in a Docker builder base image version bump too. :)


zonotope (approved):
🧮

PR #208, opened 2023-05-16
Add a test for groupBy with orderBy on a different value

cap10morgan:
Covers https://github.com/fluree/db/issues/478


dpetran (approved):
:1st_place_medal:

PR #210, opened 2023-06-06
Fix two-tuple var bind

cap10morgan:
Adds tests for two-tuple var binds (both 2 values and pred-idents) and then uses the fixed version of db in this PR: https://github.com/fluree/db/pull/496

Fixes https://github.com/fluree/db/issues/479


Jackamus29:
Just confirming this PR does fix the linked issue and one additional, un-documented issue that has to do with a change in returned data shape when using a two-tuple with a groupBy.
Thanks Wes! 🥳

test/fluree/db/ledger/query/analytical_test.clj
```diff
@@ -183,7 +183,144 @@
...
+          res    (<?? (fdb/query-async db query))]
+      (is (= [{"handle"  "jakethesnake", :_id 351843720888324
+               "favNums" [7 42], "user" {:_id 87960930223081}}]
+             res)
+          (str "Unexpected query result: " (pr-str res)))))
+  (comment
+   ;; For some reason these multiple binding vars aren't working the way I'd
+   ;; expect w/ these data and/or queries. They only return one result where I'd
+   ;; expect more than one to match. Might be a bug.
+   ;; - WSM 2023-06-06
```
zonotope, line 206: I think it makes sense to add a low priority bug in the icebox for this just so we don't forget about it if we ever need to revisit this.

zonotope (approved):

test/fluree/db/ledger/query/analytical_test.clj
cap10morgan: Issue created: https://github.com/fluree/ledger/pull/208

PR #216, opened 2023-10-23
raft file loading

dpetran:
This simplifies our file handling so that moving files around won't crash with file path errors.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -106,7 +106,7 @@ jobs:
           context: .
           file: ./Dockerfile
           platforms: linux/amd64,linux/arm64
-          push: ${{ github.event_name != 'pull_request' && steps.prep.outputs.tags != 'fluree/ledger:none' }}
+          push: ${{ steps.prep.outputs.tags != 'fluree/ledger:none' }}
           tags: ${{ steps.prep.outputs.tags }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -106,7 +106,6 @@ jobs:
           context: .
           file: ./Dockerfile
           platforms: linux/amd64,linux/arm64
-          push: ${{ github.event_name != 'pull_request' && steps.prep.outputs.tags != 'fluree/ledger:none' }}
```
cap10morgan, line 109: I don't think you want to delete this entire line. I believe that will disable pushing altogether, or if it does default to true, we still want to filter out that `fluree/ledger:none` tag.

I was just suggesting removing the `github.event_name != 'pull_request'` part.

deps.edn
```diff
@@ -2,8 +2,10 @@
             org.clojure/data.xml           {:mvn/version "0.2.0-alpha8"}
             com.fluree/alphabase           {:mvn/version "3.3.0"}
             com.fluree/db                  {:mvn/version "2.0.3"}
-            io.github.fluree/raft          {:git/sha "0fa1f90"
-                                            :git/tag "v1.0.0-beta2"}
+            io.github.fluree/raft          #_{:git/sha "0fa1f90"
+                                              :git/tag "v1.0.0-beta2"}
```
cap10morgan: We can just remove the commented-out version tag map.

src/fluree/db/server.clj
```diff
@@ -262,7 +262,7 @@
 
 (defn -main []
   (println "******************************************"
-           "\nFluree version" (meta/version)
+           "\nFluree version" (meta/version) "DEP test 2.0.4_dan-test"
```
cap10morgan: We don't want to merge that, do we?

cap10morgan (changes requested):

src/fluree/db/server.clj
dpetran: I couldn't figure out where to change the version and needed some way to tell if the tool was using the right image, but now that I know the right way I can ditch this.

cap10morgan (approved):
🎈

